PR #2, opened 2020-04-07
bence/gh actions config

underyx:
- .github: Run r2c.registry.latest via Bento Action
- bash is dumb
- update syntax again
- update syntax
- try with server
- Create CONTRIBUTING.md
- Create SECURITY.md
- update readme
- update to not check non-required args
- clean up readme
- clean up readme
- update readme
- prepare for develop
- exit properly
- try stdout
- always echo
- remove double run
- alwasy exist 0
- echo even if error
- echo even if error
- set the exitcode output
- try again without error set
- test error:false
- fix output
- update pointer
- edit syntax
- add output
- test with false
- use proper entrypoint
- stringify
- Update self_test.yml
- add error

### DIFF
--- a/.bento/config.yml
+++ b/.bento/config.yml
@@ -1,0 +1,105 @@
+autorun:
+  block: true
+formatter:
+- clippy: {}
+tools:
+  bandit:
+    ignore:
+    - any-other-function-with-shell-equals-true
+    - assert-used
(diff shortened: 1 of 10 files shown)


underyx:
wrong base branch here, see https://github.com/returntocorp/sgrep-action/pull/3

PR #4, opened 2020-05-12
Refactor repo to use bento to execute semgrep

underyx:
## Note

This removes the `output` option/feature we previously had.


CONTRIBUTING.md
```diff
@@ -1,3 +1,17 @@
 # Contributing
 
-Please fork and make a PR against `develop`. We love bugfixes and new features. Please open an issue or email semgrep@r2c.dev for help.
+## Upgrading to a new version of Bento
```
nbrahms, line 3:
```suggestion
## Upgrading to a new version of Semgrep
```

CONTRIBUTING.md
```diff
@@ -1,3 +1,17 @@
...
+
+```pipenv update```
+
+will update the version of Bento being used.
```
nbrahms, line 7:
```suggestion
will update the version of Semgrep being used.
```

entrypoint.sh
```diff
@@ -1,44 +1,115 @@
...
+  echo
+
+  git checkout "${real_head_sha}" -- .
+  git status --branch --short
+
+  echo
+  echo "== [3/3] …and seeing if there are any new findings!"
+  echo
+  bento init &> /dev/null
+  bento check --tool=semgrep
```
nbrahms, line 56: FWIW, we could now `bento check --tool=semgrep --diff-against="${GITHUB_BASE_REF}"`

entrypoint.sh
```diff
@@ -1,44 +1,115 @@
...
+    echo "== [WARNING] you didn't configure what rules semgrep should scan for."
+    echo 
+    echo "Please either set a config in the action's configuration according to"
+    echo "https://github.com/returntocorp/semgrep-action#configuration"
+    echo "or commit your own rules at the default path of .bento/semgrep.yml"
+  fi
+}
+
+main() {
+  echo "== action's environment: $($bento_path --version), $(python --version)"
```
nbrahms: Can we symlink `bento_path` so it shows up as `semgrep`?

nbrahms (commented):
Left some naming comments

CONTRIBUTING.md
underyx: These are actually correct. It's in the contributing docs (for ourselves) on how to update the bento-cli that we're using to run semgrep.

CONTRIBUTING.md
underyx: Same as https://github.com/returntocorp/semgrep-action/pull/4#discussion_r423949036

entrypoint.sh
underyx: What I did here instead was update it to print `semgrep/0.6.1, bento/0.12.0, python/3.8.1`. Is that okay?

entrypoint.sh
underyx: It somehow didn't work here, reverted https://github.com/returntocorp/semgrep/runs/668320897

underyx:
```
[ERROR] 878.9339065551758 root:util - Cannot complete running command in this git state. Unknown git ref 'develop'. To run against all diffs on this repository, please use --all.
```

README.md
```diff
@@ -1,53 +1,74 @@
...
+    - name: Semgrep
+      id: semgrep
+      uses: returntocorp/semgrep-action@v1
+      with:
+        config: r/r2c
+```
 
-The output arg `file|url`
+Note that the `r/r2c` config value
+will enable all 160+ checks from [our registry](https://semgrep.live/r).
```
pabloest: The 160+ checks will probably be out of date soon. Want to just be non-specific here and remove the number?

DrewDennison: @underyx can you also update to use `https://semgrep.live/c/r/all` I'm moving away from `r2c.` prefix as this is our community registry after a good idea from @dlukeomalley

DrewDennison: note that `r2c.` and `/c/r/r2c` sill work and get remapped server side

underyx: @pabloest I wanted to hint at the order of magnitude though, so people know what they're getting themselves into :D

How about "all the hundreds of checks"? Is that okay or still nah?

DrewDennison:
@underyx re: output. Can we wire that in or have an equivalent? I'm working on the ability to programmatically lay down these files as part of the enterprise SaaS app and I need to give the deployment config / reporting url etc. Doesn't need block this PR but just noting that. I'm fine too if config and output merge into one base url to rule them all

underyx:
> @underyx re: output. Can we wire that in or have an equivalent? I'm working on the ability to programmatically lay down these files as part of the enterprise SaaS app and I need to give the deployment config / reporting url etc. Doesn't need block this PR but just noting that. I'm fine too if config and output merge into one base url to rule them all

It would need a new bento-cli release I think :( isn't an equivalent what I'm gonna make us post to semgrep-app though as soon as this is merged? are these two outputs different in purpose somehow?

README.md
pabloest: @underyx Yep I think that works. Edit: “the hundreds of checks” instead of “all the hundreds of checks.”

DrewDennison:
@underyx they are the same :) As long as there is a way to define that endpoint in the action.yml not in a .bento/config or other file, then I can generate with the GH app that only has permission to read/write to `.github/workflows/semgrep.yml`

underyx:
> @underyx they are the same :) As long as there is a way to define that endpoint in the action.yml not in a .bento/config or other file, then I can generate with the GH app that only has permission to read/write to `.github/workflows/semgrep.yml`

Cool! That's how it's gonna work indeed.

CONTRIBUTING.md
nbrahms: 👍

entrypoint.sh
nbrahms: sure

PR #5, opened 2020-05-13
Support posting messages to semgrep-app

underyx:


action.yml
```diff
@@ -11,6 +11,12 @@ inputs:
   slackWebhookURL:
     description: |
       A Slack incoming webhook URL that you want findings posted to.
+  semgrepAppToken:
+    description: |
+      Access token to post results to the semgrep app.
```
nbrahms, line 16: In some magical future this should be pulled out of GitHub secretes

entrypoint.sh
```diff
@@ -4,13 +4,23 @@ set -euo pipefail
 bento_path=$(which bento)
```
nbrahms, line 4: At some point we should probably just make this a Python script

semgrep-monitor
```diff
@@ -50,11 +75,40 @@ def generate_message(notify_reason: str) -> Dict[str, Any]:
     ]
 
 
-def main():
+def report_start_to_semgrep_app():
+    if not args.semgrep_app_token:
+        return
```
nbrahms: Some sort of error / warning logging would be useful here

semgrep-monitor
```diff
@@ -73,13 +130,57 @@ def main():
...
-            urllib.parse.urlencode({"payload": json.dumps(payload)}).encode(),
+        response = requests.post(args.slack_url, data={"payload": json.dumps(payload)},)
+        response.raise_for_status()
+    except requests.RequestException:
+        print(f"Slack returned this error: {response.text}", file=sys.stderr)
+
+
+def report_finish_to_semgrep_app():
+    if not args.semgrep_app_token:
+        return
```
nbrahms: Ditto here

semgrep-monitor
```diff
@@ -73,13 +130,57 @@ def main():
...
+
+    # send findings
+    results = json.load(args.results_file)
+    payload = [
```
nbrahms: We may want to batch this in groups of 10000 or so to keep the payload size below 10 MB.

semgrep-monitor
```diff
@@ -73,13 +130,57 @@ def main():
...
         )
-    except urllib.error.HTTPError as e:
-        print(f"Slack returned this error: {e.read().decode()}")
+        response.raise_for_status()
+    except requests.RequestException:
+        print(f"Semgrep App returned this error: {response.text}", file=sys.stderr)
+
+    # mark as complete
+    stats = {
+        "findings": len(results)
```
nbrahms: You can put whatever stats you want here; the API example ones were just random ones I came up with in 10 seconds of brainstorming.

semgrep-monitor
nbrahms: Except for batching, we should just make `payload = results`.

There's not currently any expectation of schema on the backend.

action.yml
underyx: That's the intention. In the workflow config the user has to explicitly map their github secret to our input argument. There's no way for us to access a github secret directly.

semgrep-monitor
```diff
@@ -50,11 +75,40 @@ def generate_message(notify_reason: str) -> Dict[str, Any]:
...
 
 
-def main():
+def report_start_to_semgrep_app():
+    if not args.semgrep_app_token:
+        return
+
+    api_url = f"{args.semgrep_app_url}/api/agent/deployment/{args.deployment_id}/scan"
+    metadata = {
+        "repository": repo_name,
```
nbrahms: You should add any other relevant metadata here as well; the backend has no schema expectation.

@dlukeomalley which metadata are you looking for?

Off the top of my head, I would think that an array of committers and authors would both be useful.

semgrep-monitor
underyx: Are you sure? I expect most users would want to run semgrep-action without reporting to semgrep app. If we add a warning, they'd get that message on every run.

nbrahms (changes requested):

entrypoint.sh
underyx: Yep, that's my plan as well. Using click to allow configuration via environment variables for the whole thing would be really useful, too.

semgrep-monitor
nbrahms: Maybe hide it behind a `--verbose`?

semgrep-monitor
nbrahms: (See `bento.util.batched`)

semgrep_agent/meta.py
```diff
@@ -0,0 +1,75 @@
+import os
+
+import click
+import git
+
+from dataclasses import dataclass
+
+
+@dataclass
+class Meta:
```
nbrahms, line 11: neat

semgrep_agent/semgrep_app.py
```diff
@@ -0,0 +1,65 @@
...
+            self.scan_id = response.json()["scan"]["id"]
+
+    def report_results(self, results):
+        if self.token is None or self.deployment_id is None or self.scan_id is None:
+            return
+
+        # report findings
+        for chunk in chunked_iter(results.findings, 10_000):
+            try:
+                response = self.session.post(
```
nbrahms, line 45: 👌

semgrep_agent/bento.py
```diff
@@ -0,0 +1,115 @@
...
+def scan(ctx: click.Context) -> Results:
+    event_type = ctx.obj.event_type
+    click.echo(f"== triggered by a {event_type}")
+
+    if event_type not in ALLOWED_EVENT_TYPES:
+        fail_on_unknown_event()
+
+    bento.init()
+
+    try:
```
nbrahms, line 103: Can we add a timer around this to get the wall-clock time?

nbrahms (approved):

PR #7, opened 2020-05-14
Enrich agent with all sorts of stuff

underyx:


action.yml
```diff
@@ -17,6 +17,9 @@ inputs:
   publishDeployment:
     description: |
       Deployment ID on semgrep app.
+  generateSarif:
+    description: |
+      Set if you want a semgrep.sarif file to be written with all semgrep findings.
```
nbrahms: Probably useful to say what to set it to... e.g. `1`, `true`, etc.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+  - repo: https://github.com/psf/black
+    rev: 19.10b0
+    hooks:
+      - id: black
+
+  - repo: https://github.com/asottile/reorder_python_imports
+    rev: v2.3.0
+    hooks:
+      - id: reorder-python-imports
```
nbrahms, line 25: IMO isort >> reorder-python-imports

nbrahms (approved):

action.yml
underyx: Right. Any of them would work, but I'll say set to 1.

.pre-commit-config.yaml
underyx: I prefer it too. But don't feel like configuring it now, so just created an issue here: https://github.com/returntocorp/template/issues/7

PR #8, opened 2020-05-14
Post commit date to backend in findings

nbrahms:
This can be used to time order by commit date.


underyx (commented):
I'm actually not too happy about this one. We already sent metadata when we created the scan. That included the commit date. Why doesn't the backend work with that data?

PR #14, opened 2020-05-28
sapp: Get ignore patterns from semgrep-app

underyx:


semgrep_agent/dontfixme.py
```diff
@@ -0,0 +1,3 @@
+import pickle
+
+pickle.dumps("hi")
```
nbrahms: Did you mean to commit these?

nbrahms (approved):

semgrep_agent/dontfixme.py
underyx: I was using those files just to test the changes the PR introduced. It didn't work as I expected, was going to figure out why later.

nbrahms (approved):

underyx:
Worked here: https://github.com/returntocorp/semgrep-action/runs/721004856

PR #20, opened 2020-07-09
Move diff-awareness from bento to semgrep-agent

underyx:


nbrahms:
@underyx , the only thing that ever gave me confidence that this worked in Bento were the e2e tests (in Bento `tests/acceptance/generic*`). Thoughts on porting them here?

underyx:
Merging after having done some extensive testing.

PR #25, opened 2020-07-14
semgrep: Find current issues first and baseline second

underyx:
This lets us skip the baseline scan if there are no current issues which
could be considered new anyway.

This is the output when filtering to paths with current issues:
 
```
| using semgrep rules from r/all
| using default path ignore rules of common test and dependency directories
| looking at 122 changed files
| skipping 46 files based on path ignore rules
=== looking for current issues in 76 files
| 13 current issues found
=== looking for pre-existing issues in 5 files
| 8 pre-existing issues found
...
```

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -60,9 +60,9 @@ def main(
     publish_token: str,
     publish_deployment: int,
 ) -> NoReturn:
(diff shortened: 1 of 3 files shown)


nbrahms (approved):
I think this is reasonable. Previously we indicated how many issues you fixed, but we can do that in the backend now.

PR #28, opened 2020-07-15
Add option to allow failures

underyx:


underyx:
Not sure if we want the feature yet though.

PR #36, opened 2020-08-13
Rev to semgrep 0.19.1

nbrahms:


underyx (changes requested):
```
FROM returntocorp/semgrep:0.18.0@sha256:fa6a9a7cd2f05676bacba6f35a7f698a13a0bec2ae7b18dc7215a875f0cabd8e AS semgrep
```

This line needs to be changed as well, to bring in the correct semgrep-core.

underyx (approved):

PR #37, opened 2020-08-14
Checking finding metadata to see if should block

### LINKED ISSUE #34: Make semgrep-agent fail CI only on blocking issues
semgrep.dev will soon support different actions per rule, so

Assume that `Finding` instances can have a `"dev.semgrep.actions"` key in their `finding.metadata` dict in the `Results` object that semgrep invocations return here: https://github.com/returntocorp/semgrep-action/blob/1e1c2f06307dcdda29c6f8889c2ffb5abb88eb35/src/semgrep_agent/main.py#L120

We should make the agent follow the actions recommended by semgrep-app.

Specifically, we should exit with

```
5 notify-only findings hidden in output
<exit code 0>
``` 

and

```
| [... actual blocking errors here ...]
| [... actual blocking errors here ...]
| [... actual blocking errors here ...]
1 notify-only finding hidden in output
<exit code 1>
``` 

depending on the value of `metadata: dev.semgrep.actions: ["notify", "block"]`

### Downscoping

- We can implement this before the backend supports it. Let's just default findings to have `["block"]` as their actions when not otherwise specified.

brendongo:
Closes https://github.com/returntocorp/semgrep-action/issues/34

### DIFF
--- a/src/semgrep_agent/findings.py
+++ b/src/semgrep_agent/findings.py
@@ -46,6 +46,12 @@ class Finding:
         kw_only=True,
     )
 
+    def is_blocking(self) -> bool:
+        """
+            Returns if this finding indicates it should block CI
+        """
+        return self.metadata.get("action", "block") is "block"
+
     def syntactic_identifier_int(self) -> int:
         # Use murmur3 hash to minimize collisions
         str_id = str((self.check_id, self.path, self.syntactic_context))
--- a/src/semgrep_agent/findings.py
+++ b/src/semgrep_agent/findings.py
@@ -44,7 +44,7 @@ class Finding:
         """
             Returns if this finding indicates it should block CI
         """
-        return str(self.metadata.get("dev.semgrep.actions", "block")) == "block"
+        return "block" in self.metadata.get("dev.semgrep.actions", ["block"])
 
     def syntactic_identifier_int(self) -> int:
         # Use murmur3 hash to minimize collisions
--- a/src/semgrep_agent/findings.py
+++ b/src/semgrep_agent/findings.py
@@ -38,13 +38,13 @@ class Finding:
     commit_date = attr.ib(
         type=Optional[datetime], default=None, hash=None, eq=False, kw_only=True
     )
-    metadata = attr.ib(type=Mapping[str, Any], hash=None, eq=False, kw_only=True,)
+    metadata = attr.ib(type=Mapping[str, Any], hash=None, eq=False, kw_only=True)
 
     def is_blocking(self) -> bool:
         """
             Returns if this finding indicates it should block CI
         """
-        return str(self.metadata.get("action", "block")) == "block"
+        return str(self.metadata.get("dev.semgrep.actions", "block")) == "block"
 
     def syntactic_identifier_int(self) -> int:
         # Use murmur3 hash to minimize collisions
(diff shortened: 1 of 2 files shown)


nbrahms:
Should this be converted to a PR?

src/semgrep_agent/findings.py
```diff
@@ -38,13 +38,13 @@ class Finding:
...
-        eq=False,
-        kw_only=True,
-    )
+    metadata = attr.ib(type=Mapping[str, Any], hash=None, eq=False, kw_only=True,)
+
+    def is_blocking(self) -> bool:
+        """
+            Returns if this finding indicates it should block CI
+        """
+        return str(self.metadata.get("action", "block")) == "block"
```
underyx:
```suggestion
        return str(self.metadata.get("dev.semgrep.actions", "block")) == "block"
```

per https://github.com/returntocorp/semgrep-action/issues/34

src/semgrep_agent/findings.py
```diff
@@ -38,13 +38,13 @@ class Finding:
...
         type=Optional[datetime], default=None, hash=None, eq=False, kw_only=True
     )
-    metadata = attr.ib(
-        type=Optional[Mapping[str, Any]],
-        default=None,
-        hash=None,
-        eq=False,
-        kw_only=True,
-    )
+    metadata = attr.ib(type=Mapping[str, Any], hash=None, eq=False, kw_only=True,)
```
underyx:
```suggestion
    metadata = attr.ib(type=Mapping[str, Any], hash=None, eq=False, kw_only=True)
```

Teeny tiny change but might as well 😇

src/semgrep_agent/main.py
```diff
@@ -122,6 +122,15 @@ def main(
     formatter.dump(results.findings.new)
     obj.sapp.report_results(results)
 
-    exit_code = 1 if results.findings.new else 0
+    exit_code = get_error_code(results)
     click.echo(f"=== exiting with {'failing' if exit_code == 1 else 'success'} status")
```
underyx, line 138: Could you also print the number of results that were not printed? Something like the sample output in https://github.com/returntocorp/semgrep-action/issues/34#issue-676927729

underyx (changes requested):

brendongo:
IMO much easier to understand what's happening by doing the set operations inline for now rather than having Results/FindingSet expose a bunch of one line functions. Thoughts?

underyx:
> IMO much easier to understand what's happening by doing the set operations inline for now rather than having Results/FindingSet expose a bunch of one line functions. Thoughts?

Yes, I'm a fan of inline over one-line functions always and forever! 🎉

src/semgrep_agent/findings.py
```diff
@@ -38,13 +38,13 @@ class Finding:
...
-        eq=False,
-        kw_only=True,
-    )
+    metadata = attr.ib(type=Mapping[str, Any], hash=None, eq=False, kw_only=True)
+
+    def is_blocking(self) -> bool:
+        """
+            Returns if this finding indicates it should block CI
+        """
+        return str(self.metadata.get("dev.semgrep.actions", "block")) == "block"
```
underyx:
```suggestion
        return "block" in self.metadata.get("dev.semgrep.actions", ["block"])
```

I missed this earlier, but the metadata field is an array of actions, so we need to handle it as such

PR #40, opened 2020-08-19
Use file targeting instead of include when invoking semgrep

### LINKED ISSUE #35: Agent takes longer than running semgrep directly
Looks like semgrep-agent is passing files to semgrep using --includes which causes semgrep to traverse the directory tree to see if any files match the include pattern.

Fix will be to pass the files directly to semgrep as a target argument.

brendongo:
With --skip-unknown-extension flag in semgrep 0.20.0 we can pass
explicit files directly to semgrep without running into issue of
semgrep running all explicitly passed files on all rules.

This makes semgrep aaction run as fast as semgrep locally since it
no longer needs to traverse the file tree with the --include passing

Closes https://github.com/returntocorp/semgrep-action/issues/35


underyx (approved):
☺️ feel free to release according to CONTRIBUTING.md after merging!

PR #46, opened 2020-09-01
Use file targeting instead of include for checking pre-existing issues

Isaac0616:
Similar to 11ee1a6f37fbf232b3f39afc97fce12f4b854057, we should also use file targeting for checking pre-existing issues to improve performance.


Isaac0616:
Question: Does [this line](https://github.com/returntocorp/semgrep-action/blob/develop/src/semgrep_agent/semgrep.py#L143)  suppose to live within the for chunk loop?

brendongo (approved):

brendongo:
CI doesn't play well with outside contributors yet. Will look into in a bit

brendongo:
@Isaac0616 I'm going to close this PR and reopen one myself so tests run

brendongo:
Moved to https://github.com/returntocorp/semgrep-action/pull/49

PR #48, opened 2020-09-02
output: only findings to stdout and add json output option

### LINKED ISSUE #42: Add the ability to configure Semgrep core via the agent
Currently, the agent can only accept the config file and prints out the results to stdout (absent other switches). There's no way to get a json output to process the results using the command line.

It would be nice to be able to configure the Semgrep core that is running inside the agent. Mainly, the ability to get the JSON output. I think the easiest way to do so is by accepting Semgrep code switches (e.g., `--json`). This will make the transition between running Semgrep and the agent seamless. The agent is the suggested way of running Semgrep in a CI pipeline so having access to the json output in the container is great.

It seems like this is the location where the context (that contains the config) to Semgrep.

https://github.com/returntocorp/semgrep-action/blob/develop/src/semgrep_agent/main.py#L121

brendongo:
Hidden simple flag for json output. Can consider other semgrep flags and interaction with semgrep through semgrep-agent in the future

### DIFF
--- a/src/semgrep_agent/formatter.py
+++ b/src/semgrep_agent/formatter.py
@@ -95,4 +95,4 @@ def dump(findings: Set[Finding]) -> None:
             lines.append("")
 
     for line in lines:
-        click.secho(line)
+        click.secho(line, err=True)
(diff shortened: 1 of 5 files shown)


underyx (approved):
Oh thank you, I've been meaning to go through all those echo calls 😌

dlukeomalley:
@brendongo @underyx Should we have a semgrep rule to scan for usage of echo?

PR #52, opened 2020-09-08
Add attrs as dependency

### LINKED ISSUE #51: Dependencies of semgrep-agent
Maybe should add the `attrs` package to https://github.com/returntocorp/semgrep-action/blob/develop/pyproject.toml#L11 since `semgrep-agent` uses it directly?

brendongo:
Newer version 20.2.0 exists and seems to work but keeping ~=19.3.0 for now to match semgrep dependency


brendongo:
Closes https://github.com/returntocorp/semgrep-action/issues/51

PR #53, opened 2020-09-08
Remove .semgrepignore support

brendongo:
.semgrepignore was a stop gap to get file ignores while
semgrep did not yet have excludes/includes properly implemented

Now that semgrep has this functionality it is fine to remove here
so we don't have to support it in different places. Also the
include/exclude in semgrep is much more efficient since it does
not need to walk the entire file tree.


src/semgrep_agent/semgrep.py
```diff
@@ -65,33 +65,6 @@ def get_semgrep_config(ctx: click.Context) -> Iterator[List[Union[str, Path]]]:
...
-        semgrepignore.write((TEMPLATES_DIR / ".semgrepignore").read_text())
-
-    scan_patterns = scan.ignore_patterns
-    if scan_patterns:
-        click.echo(
-            "| adding further path ignore rules configured on the web UI", err=True
-        )
-        semgrepignore.write("\n# Ignores from semgrep app\n")
-        semgrepignore.write("\n".join(scan_patterns))
-        semgrepignore.write("\n")
```
underyx, line 90: This is a semgrep.dev feature that we still intend to support. We need some sort of replacement for using the file ignores that are configured on the web UI.

underyx (changes requested):
We can't break the UI-based file ignore config :/ Can we still pass those paths to semgrep CLI?

src/semgrep_agent/semgrep.py
brendongo: Oh I see. Good call.

brendongo:
Closing for now since this is not high priority and base has already shifted under it.

PR #54, opened 2020-09-09
refactor: explicitly pass args instead of passing ctx object

brendongo:
Using ctx object as a global made it hard to reason about behavior.

This PR explicitly passes necessary arguments to function instead of passing
the ctx object as a whole.


brendongo:
After this https://github.com/returntocorp/semgrep-app/issues/515 is simple

underyx (approved):
That's some impressing unraveling work. Nice!

PR #62, opened 2020-09-23
Add finding index to findings hash

### LINKED ISSUE #56: Diff-aware doesn't report new findings if the new finding has the same (rule_id, path, line_of_code) as the one before
Assume you have this diff:

```diff
  eval(foo)
  # some code
+ eval(foo)
```

The addition of a new eval security issue should be warned about on a PR. Right now we use a `set` to figure out what findings are new, which would make the agent unaware about the new issue. If we used a `Counter`, we could warn that there are more instances of the same thing now.

nbrahms:
In order to not trigger a new finding on whitespace changes and line
moves, we exclude surrounding context data from findings hashes.
However, this means that duplicate results with the exact same code in
a file will show up as only one finding, which is not desirable.

Here, we include the index of a finding within the file, so these show
up as separate findings.

This was tested by committing a "bad.py" with:
```
  5 == 5
```
Then committing an edit to that file to read:
```
  5 == 5

  5 == 5
```
We validated that the agent found 2 findings in the second commit, and
1 new finding when comparing the second commit to the first.

I then validated the finding number sent to the backend with:
```
    semgrep-agent \
      --config r/python.lang.correctness.useless-eqeq.useless-eqeq \
      --baseline-ref head^^ \
      --publish-url "https://dev.semgrep.dev" \
      --publish-token '********' \
      --publish-deployment 1
```

Fixes #56

### DIFF
--- a/src/semgrep_agent/findings.py
+++ b/src/semgrep_agent/findings.py
@@ -7,13 +7,19 @@ from datetime import datetime
 from typing import Any
 from typing import Dict
 from typing import Mapping
-from typing import NamedTuple
+from typing import List
 from typing import Optional
 from typing import Set
 
 import attr
-import click
 import pymmh3
+
+
+@attr.s(frozen=True)
+class FindingKey:
+    check_id = attr.ib(type=str)
+    path = attr.ib(type=int)
+    syntactic_context = attr.ib(type=str)
 
 
 @attr.s(frozen=True, hash=False)
@@ -28,6 +34,7 @@ class Finding:
     column = attr.ib(type=int, hash=None, eq=False)
     message = attr.ib(type=str, hash=None, eq=False)
     severity = attr.ib(type=int, hash=None, eq=False)
+    index = attr.ib(type=int, hash=None, eq=False)
     syntactic_context = attr.ib(type=str, converter=textwrap.dedent)
     end_line = attr.ib(
         type=Optional[int], default=None, hash=None, eq=False, kw_only=True
@@ -48,7 +55,7 @@ class Finding:
 
     def syntactic_identifier_int(self) -> int:
         # Use murmur3 hash to minimize collisions
-        str_id = str((self.check_id, self.path, self.syntactic_context))
+        str_id = str((self.check_id, self.path, self.index, self.syntactic_context))
         return pymmh3.hash128(str_id)
 
     def syntactic_identifier_str(self) -> str:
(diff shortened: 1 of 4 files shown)


src/semgrep_agent/main.py
```diff
@@ -126,7 +126,7 @@ def main(
         meta.base_commit_ref,
         semgrep.get_semgrepignore(sapp.scan.ignore_patterns),
     )
-    new_findings = results.findings.new
+    new_findings = results.findings.expensive_new()
```
brendongo: Why not just results.new?

brendongo (commented):
Would overriding hash of Finding to only use check_id, path, and syntactic_context do what you need? If possible would like to avoid all the Dict <-> Set overhead.

nbrahms:
> Would overriding hash of Finding to only use check_id, path, and syntactic_context do what you need? If possible would like to avoid all the Dict <-> Set overhead.

Alas, no. Two findings with the same code in the same file would show up with the same hash in that case.

src/semgrep_agent/main.py
nbrahms: good catch.

src/semgrep_agent/semgrep.py
```diff
@@ -78,15 +80,19 @@ def get_semgrepignore(ignore_patterns: List[str]) -> TextIO:
...
-    findings: FindingSets
-    total_time: float
+@attr.s(frozen=True)
+class Results(object):
+    findings = attr.ib(type=FindingSets)
+    total_time = attr.ib(type=float)
+    new = attr.ib(type=set, init=False)
+
+    def __attrs_post_init__(self) -> None:
+        object.__setattr__(self, "new", self.findings.expensive_new())
```
underyx, line 91: Could we add a comment explaining why `self.new = self.findings.expensive_new()` wouldn't work? 😕

src/semgrep_agent/findings.py
```diff
@@ -99,9 +116,22 @@ def to_dict(self) -> Mapping[str, Any]:
...
+    def expensive_new(self) -> Set[Finding]:
+        return self.current_set() - self.baseline_set()
+
+    @staticmethod
+    def _map_to_set(mapping: Mapping[FindingKey, List[Finding]]) -> Set[Finding]:
+        return set(
+            attr.evolve(f, index=ix)
+            for ff in mapping.values()
+            for ix, f in enumerate(ff)
+        )
```
underyx, line 131:
```python
        return {
            attr.evolve(finding, index=index)
            for findings in mapping.values()
            for index, finding in enumerate(findings)
        }
```

Wrote this out for myself to understand what's going on, might as well update the code as well 🙃

nbrahms: aye aye

src/semgrep_agent/findings.py
```diff
@@ -99,9 +116,22 @@ def to_dict(self) -> Mapping[str, Any]:
 
 @dataclass(frozen=True)
 class FindingSets:
-    baseline: Set[Finding] = field(default_factory=set)
-    current: Set[Finding] = field(default_factory=set)
+    baseline_map: Dict[FindingKey, List[Finding]] = field(default_factory=dict)
+    current_map: Dict[FindingKey, List[Finding]] = field(default_factory=dict)
+
+    def expensive_new(self) -> Set[Finding]:
+        return self.current_set() - self.baseline_set()
```
underyx, line 123: The index-based deduplication here is going to be yielding some wrong output. See this diff:


```diff
- eval("pre-vetted, safe string")
+ eval(request.GET.foo)
  # here comes the dangerous code, bail before running it in prod
  if not app.DEBUG:
      return
  eval(request.GET.foo)
```

The semgrep-agent would now output line 5 as the new finding instead of line 1, right? Cause its index changed from 0 to 1, and index 1 is the never-before seen index. And a human code reviewer might dismiss that as a false positive looking at the debug mode gate.

So perhaps we'll need to look at the git diff to figure out what to report as new. Should we just ship this version now anyway since it's a bit better than what we have today?

brendongo:
> Alas, no. Two findings with the same code in the same file would show up with the same hash in that case.

Doesn't this PR use those 3 fields as a key in the dict. So doesn't this mean same code in the same file also get lost?

src/semgrep_agent/findings.py
nbrahms: Short answer is that I don't see a good technical solution to this problem without really understanding the code that's written, as distinguishing between a line move and a delete + add is nigh impossible.

So I'd go with just leaving this as a wart on the solution.

underyx: If we allow this limitation, wouldn't this same thing be accomplished with the few-line change of making the original `findings.current.update(Finding.from_semgrep_result(...))` line we have be locally aware of the `index` of the result, and then including the index in the hash? I'm thinking this, in pseudocode:

```python
finding = Finding.from_semgrep_result(...)
while finding in findings.current:
    finding.index += 1
findings.current.add(finding)
```

PR #63, opened 2020-09-24
Add finding index to findings hash, try 2

underyx:


src/semgrep_agent/semgrep.py
```diff
@@ -153,25 +136,24 @@ def invoke_semgrep(
             args = ["--skip-unknown-extensions", "--json", *config_args]
             for path in chunk:
                 args.append(path)
-            count = 0
             for result in json.loads(str(semgrep(*args)))["results"]:
-                _update_finding_set(result, committed_datetime, findingSet.current_map)
-                count += 1
+                finding = Finding.from_semgrep_result(result, committed_datetime)
+                while finding in findings.current:
```
nbrahms: I think this won't work under the following test plan:

Commit:
```py
5 == 5
```

Then alter file to
```py
6 == 6

5 == 5
```

You should only see 1 new finding in this case.

nbrahms (changes requested):

src/semgrep_agent/semgrep.py
nbrahms: FWIW I ran this test plan on the existing code, and it works as it should currently.

nbrahms: OK, thinking about this a bit more, I think ^^ will pass.

However, I see issues with the following:

Commit a new file with two findings in it:
```
2 == 2

2 == 2
```
The agent will only send one finding to the backend, and it will have the hash with index = 1.

Then if we fix one of the findings:
```
2 == 2
```
The backend will detect a new finding (!), because it now has the hash with index 0, for the first time.

nbrahms: Also just tested this, and your code works. Took me a while to understand why, so maybe that's a flag that a comment is needed to explain your algorithm.

But your solution is definitely simpler than mine, so let's use it.

Also FYI, @brendongo expressed a preference for putting the finding-set mutation logic inside the FindingsSet object, so may want him to check out this code too.

nbrahms (approved):

src/semgrep_agent/semgrep.py
brendongo: Yes if possible it would be nice if users of the FindingsSet object don't need to know about indexes and having to evolve finding objects. Can probably have FindingSets expose an add_to_baseline/current method

brendongo (changes requested):
See comment re: not exposing needing to update Finding Index when adding a Finding to a FindingsSets by exposing a helper method in FindingSets.

Nice work. Should we revert the other PR or do you want to just handle fixing merge conflics?

underyx:
Agree on merging this code into `FindingSet`, I just wanted this to go through review before abstracting it.

As for the merge conflicts, not sure what happened, this PR reverts the previous merge with its first commit: https://github.com/returntocorp/semgrep-action/pull/63/commits/66a16c0e30d4f53639a16675eed6c1377c682199 😕 

Maybe I did the revert wrong, will fix tomorrow!

nbrahms:
> Agree on merging this code into `FindingSet`, I just wanted this to go through review before abstracting it.
> 
> As for the merge conflicts, not sure what happened, this PR reverts the previous merge with its first commit: [66a16c0](https://github.com/returntocorp/semgrep-action/commit/66a16c0e30d4f53639a16675eed6c1377c682199) 😕
> 
> Maybe I did the revert wrong, will fix tomorrow!

It's because I did some refactoring after the previous PR. I'd say just rebase and fix the conflicts?

underyx:
@brendongo please see https://github.com/returntocorp/semgrep-action/pull/63/commits/94111797734aaf520f13f4e5c5b1351531b415ea — it refactors this so that the only change compared to the original semgrep results consumption code is going from `findings.current.update(...)` to `findings.current.update_findings(...)`. This `update_findings` method is implemented on a set subclass.

brendongo (approved):

PR #70, opened 2020-09-25
Set SEMGREP_USER_AGENT_APPEND to distinguish environment

mschwager:
For https://github.com/returntocorp/enterprise/issues/359.

Is there a good way to test that this is working?

### DIFF
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -37,7 +37,7 @@ from semgrep_agent.utils import debug_echo
 if TYPE_CHECKING:
     from semgrep_agent.semgrep_app import Scan
 
-ua_environ = {**os.environ, **{'SEMGREP_USER_AGENT_APPEND': '(Agent)'}}
+ua_environ = {"SEMGREP_USER_AGENT_APPEND": "(Agent)", **os.environ}
 semgrep = sh.semgrep.bake(_ok_code={0, 1, 2}, _tty_out=False, _env=ua_environ)
 
 # a typical old system has 128 * 1024 as their max command length
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -37,7 +37,7 @@ from semgrep_agent.utils import debug_echo
 if TYPE_CHECKING:
     from semgrep_agent.semgrep_app import Scan
 
-ua_environ = {**os.environ, **{'SEMGREP_USER_AGENT_APPEND', '(Agent)'}}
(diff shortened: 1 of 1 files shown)


src/semgrep_agent/semgrep.py
```diff
@@ -37,7 +37,8 @@
 if TYPE_CHECKING:
     from semgrep_agent.semgrep_app import Scan
 
-semgrep = sh.semgrep.bake(_ok_code={0, 1, 2}, _tty_out=False)
+ua_environ = {**os.environ, **{'SEMGREP_USER_AGENT_APPEND', '(Agent)'}}
```
underyx:
```suggestion
ua_environ = {**os.environ, **{'SEMGREP_USER_AGENT_APPEND': '(Agent)'}}
```
Did you mean this?

mschwager: Yes, good catch!

underyx:
```suggestion
ua_environ = {"SEMGREP_USER_AGENT_APPEND": "(Agent)", **os.environ}
```

This should be nicer though anyway. Also allows overriding with an env var visible to semgrep-agent if you really want to.

src/semgrep_agent/semgrep.py
```diff
@@ -37,7 +37,8 @@
 if TYPE_CHECKING:
     from semgrep_agent.semgrep_app import Scan
 
-semgrep = sh.semgrep.bake(_ok_code={0, 1, 2}, _tty_out=False)
+ua_environ = {**os.environ, **{'SEMGREP_USER_AGENT_APPEND': '(Agent)'}}
```
underyx:
```suggestion
ua_environ = {"SEMGREP_USER_AGENT_APPEND": "(Agent)", **os.environ}
```

This should be nicer though anyway. Also allows overriding with an env var visible to semgrep-agent if you really want to.

underyx: Sorry, I didn't write this comment in time previously 😅

underyx:
As for testing, I'm not sure — perhaps it's enough to see if `(Agent)` user agents start showing up in our server logs and analytics?

mschwager:
@DrewDennison mentioned we could check live analytics via Segment. Would the CI test run with this new functionality show up there for us to confirm?

mschwager:
Confirmed in Segment: `"Semgrep/0.25.0 (Agent)"` we should be good to go.

underyx (approved):

Issue #61, opened 2020-09-21
Add support for adding a number of yml rules from the .semgrep folder 


tkisason:
The current semgrep-action can rely on either
1. The online rule registry
2. a `.semgrep.yml` file with a collection of rules, which can be unwieldily to manage for a large number of rules.

The "normal" semgrep, supports a `.semgrep` folder, that can contain a number of rules in the  `.semgrep/**/*.yml` form, where it is easy to mantain a number of semgrep rules in a single folder for a project. 

So, if it is possible, i would suggest a feature, where semgrep-action would support the same method for rule passing, as this would make it easy to manage larger rulesets for CI (where a project can contain it's rulefolder), where a external registry can't be used because of regulatory requirements. 

Thanks!

tkisason:
Thank you!

PR #75, opened 2020-09-29
Update README.md with info about .semgrep

tkisason:
Update README.md with info about the .semgrep folder changes from https://github.com/returntocorp/semgrep-action/issues/61


underyx (approved):
Thank you for all the help!

PR #76, opened 2020-09-29
Calculate merge base in GHA

### LINKED ISSUE #73: Action reports issues on trunk branch
Action appears to be reporting issues that occur in PR _as well as_ any new issues on the merge target (e.g. `master`).

It should only report issues in the PR itself.

Suggestion:
Calculate `--diff-against` using `git merge-base` of PR commit and target branch.

nbrahms:
Fixes #73

### DIFF
--- a/src/semgrep_agent/meta.py
+++ b/src/semgrep_agent/meta.py
@@ -123,7 +123,10 @@ class GithubMeta(GitMeta):
     @cachedproperty
     def base_commit_ref(self) -> Optional[str]:
         if self.event_name == "pull_request":
-            return self.glom_event(T["pull_request"]["base"]["sha"])  # type: ignore
+            pr_base = self.glom_event(T["pull_request"]["base"]["sha"])
+            # The pull request "base" that GitHub sends us is not necessarily the merge base,
+            # so we need to get the merge-base from Git
+            return git("merge-base", pr_base, "HEAD").stdout.decode().strip()
         return None
 
     @cachedproperty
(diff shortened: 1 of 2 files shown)


src/semgrep_agent/semgrep.py
```diff
@@ -29,7 +29,6 @@
 
 from semgrep_agent.findings import Finding
 from semgrep_agent.findings import FindingSets
-from semgrep_agent.meta import GitMeta
```
nbrahms, line 32: this was unused

src/semgrep_agent/meta.py
```diff
@@ -123,7 +123,10 @@ def commit_sha(self) -> Optional[str]:
     @cachedproperty
     def base_commit_ref(self) -> Optional[str]:
         if self.event_name == "pull_request":
-            return self.glom_event(T["pull_request"]["base"]["sha"])  # type: ignore
+            pr_base = self.glom_event(T["pull_request"]["base"]["sha"])
+            # The pull request "base" that GitHub sends us is not necessarily the merge base,
+            # so we need to get the merge-base from Git
+            return git("merge-base", pr_base, "HEAD").stdout.decode().strip()
```
brendongo, line 129: Is this specific to Github or will Gitlab base_commit_ref face the same issue?

brendongo (approved):

src/semgrep_agent/meta.py
nbrahms: It looks like GitLab already does something with `merge-base`, although I don't understand what `fetch-head` is as a ref, so I didn't monkey with it.

nbrahms: (see line 209)

PR #83, opened 2020-10-07
Create PRIVACY.md

sabrinabrogren:


PRIVACY.md
```diff
@@ -0,0 +1,21 @@
+# Semgrep Privacy Policy
+
+Semgrep collects usage data to help us improve the product. We send two types of data to r2c servers: scan data and findings data.
+
+## Scan data
+
+Scan data provide information on the environment and performance of Semgrep. They power dashboards, identify anomalies with the product, and are needed for billing. The classes of data included are:
+- Project identity (e.g. name, URL)
+- Scan environment (e.g. CI provider, OS)
+- Author identity (e.g. committer email)
```
dlukeomalley, line 10: To avoid PII, is it possible to set this to be a hash of committer email?

PRIVACY.md
```diff
@@ -0,0 +1,21 @@
...
+- Commit metadata (e.g. commit hash)
+- Scan metadata, including type of scan and scan parameters (e.g. files or branches to ignore)
```
dlukeomalley, line 12: What are some examples of this data? Is file information needed? See later comments on not including file path in Findings data.

dlukeomalley: Would this data include timing information?

PRIVACY.md
```diff
@@ -0,0 +1,21 @@
...
+- Review and review-requester identifying data (e.g pull-request ID, branch, merge base, request author)
```
dlukeomalley, line 13: Request author would also be PII. Can we include it as a hash for now?

PRIVACY.md
```diff
@@ -0,0 +1,21 @@
...
+- Semgrep environment (e.g. version, interpreter, timestamp)
+
```
dlukeomalley, line 15: Is IP address or client information collected as part of scan data?

PRIVACY.md
```diff
@@ -0,0 +1,21 @@
...
+## Findings data
+
+Findings data are used to provide human readable content for notifications and integrations, as well tracking results as new, fixed, or duplicate. The classes of data included are:
+- Check ID and metadata (as defined in the rule definition; e.g. OWASP category, message, severity)
```
dlukeomalley, line 19: Message, in the case of metavariables, could include sensitive data. Is it required and where/when would it be used? Is the other data, such as OWASP category and severity something we can infer from the rule-id rather than sending with every finding? How would we handle rules that are only local?

PRIVACY.md
```diff
@@ -0,0 +1,21 @@
...
+- Code location (including file path) that triggered finding, **but not code content**
```
dlukeomalley: File path is considered sensitive by some. What does this data get used for and could it be "opt-in" if it's required for notifications?

dlukeomalley (changes requested):
@sabihb @DrewDennison @nbrahms Requesting changes, along with the relevant BE changes, to limit the data we collect to avoid PII and fields that are viewed as sensitive by some (highly dependent on an individuals threat model).

PRIVACY.md
```diff
@@ -0,0 +1,21 @@
+# Semgrep Privacy Policy
+
+Semgrep collects usage data to help us improve the product. We send two types of data to r2c servers: scan data and findings data.
```
daghan: Semgrep collects usage data to provide useful results and to help us improve the product. We send two types of data to r2c servers: scan data and findings data.

daghan (commented):
We should mention that we also collect IP address

PRIVACY.md
DrewDennison: yes hashing with a salt is what we did in Bento and that worked well

PRIVACY.md
DrewDennison: yes +1

PRIVACY.md
DrewDennison: We don't directly capture it but it will show up in load balancer logs due to the nature of connections

PRIVACY.md
DrewDennison: I don't think we need the message unless it's going to be pushed through to slack which some users might want to opt-in for

PRIVACY.md
DrewDennison: slack and other integration data

PRIVACY.md
DrewDennison: yes aggregate timing and performance data is collected so we can debug and improve semgrep performance proactively

PRIVACY.md
daghan: opt-in sounds great

PRIVACY.md
nbrahms: We currently use the author identity to render the Slack Webhook. If we want to explicitly ban the collection of author identities, we'll need to reconsider our integrations.

Note that collection != storage, so we can delineate between what we transmit and what we store.

PRIVACY.md
nbrahms: See comment above re. rendering webhooks.

PRIVACY.md
sabrinabrogren: Bence posted some examples he pulled from the db in [this issue](https://github.com/returntocorp/semgrep-app/issues/502).

PRIVACY.md
daghan: +1 on _Note that collection != storage, so we can delineate between what we transmit and what we store._

PRIVACY.md
daghan: I saw them and added a positive comment

PRIVACY.md
daghan: I don't think that's a big deal (based on our call with the lawyers)

nbrahms:
> We should mention that we also collect IP address

I don't think we "collect" IP addresses in a way that's exceptional from any other web-app on Earth, so I'd leave it out of this document.

nbrahms (approved):

dlukeomalley (approved):
Approving. Thanks for the depth of discussion and refinement everyone!

Issue #84, opened 2020-10-07
semgrep failed


sfc-gh-ddong:
```
Run ./.github/actions/semgrep-action
/usr/bin/docker run --name returntocorpsemgrepactionv1_7098e9 --label 1e5c35 --workdir /github/workspace --rm -e INPUT_CONFIG -e INPUT_PUBLISHTOKEN -e INPUT_PUBLISHDEPLOYMENT -e INPUT_GENERATESARIF -e HOME -e GITHUB_JOB -e GITHUB_REF -e GITHUB_SHA -e GITHUB_REPOSITORY -e GITHUB_REPOSITORY_OWNER -e GITHUB_RUN_ID -e GITHUB_RUN_NUMBER -e GITHUB_RETENTION_DAYS -e GITHUB_ACTOR -e GITHUB_WORKFLOW -e GITHUB_HEAD_REF -e GITHUB_BASE_REF -e GITHUB_EVENT_NAME -e GITHUB_SERVER_URL -e GITHUB_API_URL -e GITHUB_GRAPHQL_URL -e GITHUB_WORKSPACE -e GITHUB_ACTION -e GITHUB_EVENT_PATH -e GITHUB_PATH -e GITHUB_ENV -e RUNNER_OS -e RUNNER_TOOL_CACHE -e RUNNER_TEMP -e RUNNER_WORKSPACE -e ACTIONS_RUNTIME_URL -e ACTIONS_RUNTIME_TOKEN -e ACTIONS_CACHE_URL -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/runner/work/_temp/_github_home":"/github/home" -v "/home/runner/work/_temp/_github_workflow":"/github/workflow" -v "/home/runner/work/_temp/_runner_file_commands":"/github/workspace" returntocorp/semgrep-action:v1
Unable to find image 'returntocorp/semgrep-action:v1' locally
v1: Pulling from returntocorp/semgrep-action
df20fa9351a1: Pulling fs layer
36b3adc4ff6f: Pulling fs layer
4db9de03f499: Pulling fs layer
cd38a04a61f4: Pulling fs layer
9a3838385f13: Pulling fs layer
09359e37df4b: Pulling fs layer
2593afa0e612: Pulling fs layer
cff1f9ba2a6e: Pulling fs layer
27800508e272: Pulling fs layer
84c0aae16fc3: Pulling fs layer
[...]

PR #85, opened 2020-10-07
fix: Only call git rm if there are files to remove

brendongo:
Was failing since git rm was being called when there were no files to revert


brendongo:
Fixes https://github.com/returntocorp/semgrep-action/issues/84

colleend (approved):
lgtm

PR #87, opened 2020-10-14
Linking to a page for more details

ievans:


ievans:
@underyx CI failures seem unrelated?

src/semgrep_agent/formatter.py
```diff
@@ -88,7 +164,11 @@ def dump(findings: Set[Finding]) -> None:
 
     for path, vv in itertools.groupby(ordered, path_of):
         for v in sorted(vv, key=lambda v: (v.line, v.column, v.message)):
```
underyx, line 166: vvvvv confusing variable names here :D

src/semgrep_agent/formatter.py
```diff
@@ -68,9 +84,69 @@ def _print_error_message(violation: Finding) -> str:
...
+    # Coloring has to occur after justification
+    if is_rendered:
+        text = style(text, fg=Colors.LINK)
+
+    return text
+
+
+def build_action_url(v: Finding, c: Sapp, meta: GitMeta) -> str:
+    projectName = meta.repo_remote_origin
+    return f"http://localhost:3000/finding/{c.deployment_id}/{v.from_policy_id}?ruleId={v.check_id}&path={v.path}&lineNum={v.line}&rulesetName={v.from_ruleset_id}&projectName={projectName}&policyName={v.from_policy_name}"
```
underyx, line 144: I think the Sapp variable has the actual remote URL as an attribute.

underyx: Also needs URL encoding; ampersands for instance are probably legal in ruleset names, etc.

src/semgrep_agent/formatter.py
```diff
@@ -68,9 +84,69 @@ def _print_error_message(violation: Finding) -> str:
...
+
+
+def is_child_process_of(pattern: Pattern) -> bool:
+    """
+    Returns true iff this process is a child process of a process whose name matches pattern
+    """
+    me = psutil.Process()
+    parents = me.parents()
+    matches = iter(0 for p in parents if pattern.search(p.name()))
+    return next(matches, None) is not None
```
underyx, line 98: While it works, `next(iter(...))` is kinda unpythonic. How about

```py
return any(pattern.search(parent.name) for parent in me.parents())
```

?

src/semgrep_agent/formatter.py
```diff
@@ -68,9 +84,69 @@ def _print_error_message(violation: Finding) -> str:
...
 
 
-def dump(findings: Set[Finding]) -> None:
+OSC_8 = "\x1b]8;;"
+BEL = "\x07"
+
+
+def is_child_process_of(pattern: Pattern) -> bool:
+    """
+    Returns true iff this process is a child process of a process whose name matches pattern
```
underyx, line 93:
```suggestion
    Returns true iff this process is a child process of a process whose name contains pattern
```

(matches implies `re.match`-like behavior)

src/semgrep_agent/formatter.py
```diff
@@ -68,9 +84,69 @@ def _print_error_message(violation: Finding) -> str:
...
+
+
+LINK_PRINTER_PATTERN = re.compile("(iterm2|gnome-terminal)", re.IGNORECASE)
+LINK_WIDTH = 2 * len(OSC_8) + 2 * len(BEL)
+DO_PRINT_LINKS = is_child_process_of(LINK_PRINTER_PATTERN)
+
+
+def render_link(
+    text: str,
+    href: Optional[str],
```
underyx, line 108:
```suggestion
    href: Optional[str],
    *,
```

I'd force setting the param name for everything below, too easy to mix up the order of boolean flags.

src/semgrep_agent/meta.py
```diff
@@ -43,6 +44,11 @@ def repo(self) -> gitpython.Repo:  # type: ignore
     def repo_name(self) -> str:
         return Path.cwd().name
 
+    @cachedproperty
+    def repo_remote_origin(self) -> str:
+        remotes: List[str] = [remote.url for remote in self.repo.remotes]
```
underyx, line 49:
```suggestion
        remotes: List[Any] = sorted(self.repo.remotes, key=lambda x: x.remote == "origin")
        return remotes[-1].url if remotes else None
```

1. Other methods return `None` when no answer is available
2. This prioritizes `origin` if available

src/semgrep_agent/formatter.py
```diff
@@ -68,9 +84,69 @@ def _print_error_message(violation: Finding) -> str:
...
+
+    # Coloring has to occur after justification
+    if is_rendered:
+        text = style(text, fg=Colors.LINK)
+
+    return text
+
+
+def build_action_url(v: Finding, c: Sapp, meta: GitMeta) -> str:
+    projectName = meta.repo_remote_origin
```
underyx, line 143:
```suggestion
    projectName = meta.repo_name
```

Isn't this the correct name if we want this to identify the configurable project?

underyx (changes requested):
Scrutinized this one a bit more than usual since it's hackathon code, hope that's alright.

src/semgrep_agent/formatter.py
ievans: Good idea and willdo (By the way -- this function is copied straight from the old bento code :))

dlukeomalley:
Closing for now. Thank you for the contribution, and please re-open if this was in error!

PR #91, opened 2020-10-16
Inline PR comments

sabrinabrogren:


nbrahms (approved):

Gby56:
Hi ! I'm trying to understand where the action/container is now sending inline comments to PRs ? (I'm using the container in Jenkins for semgrep-agent)
I tried finding this string "Semgrep was unable to comment inline on the following findings" in the r2c org but no luck, and I don't understand why it's failiing to find the correct file + line number

Gby56:
Ohhhh nevermind @sabrinabrogren , I think I understand the flow, semgrep agent reports to semgrep app, and semgrep app is the one actually commenting to Github, I forgot !
Any change we could understand more why it fails to comment ? it's unclear, I think we gave the proper permissions, and `Semgrep was unable to comment inline on the following findings` is just maybe a bit vague 🤔

raghavjain3:
@Gby56 absolutely! I filed a ticket for this internally as one of our projects for the next few months will be improving the PR comment experience (including better error messaging)

PR #102, opened 2020-11-06
Surface environment variables for generic CI environments

chmccreery:
Surface environment variables to make notifications in generic CI environments possible:

- SEMGREP_REPO_URL=“https://github.com/my_repo” is used to create links to your repo files from notifications. Not sure these actually work if they are not GitHub, as specific paths may be constructed differently in different CI providers
- SEMGREP_JOB_URL=“https://github.com/myjob” is used to link to the job that produced the result
- SEMGREP_BRANCH=“mybranch” is used in the name of the notification (overrides the auto-detected branch)


src/semgrep_agent/meta.py
```diff
@@ -35,13 +35,18 @@ def event_name(self) -> str:
...
     def repo(self) -> gitpython.Repo:  # type: ignore
-        repo = gitpython.Repo()
+        repo = gitpython.Repo(".", search_parent_directories=True)
         debug_echo(f"found repo: {repo!r}")
         return repo
 
     @cachedproperty
     def repo_name(self) -> str:
-        return Path.cwd().name
+        repo = gitpython.Repo(".", search_parent_directories=True)
```
brendongo: Do we want to traverse up or should we just let users always call semgrep action at repo root?

src/semgrep_agent/meta.py
```diff
@@ -223,21 +252,16 @@ def event_name(self) -> str:
...
-    def mr_id(self) -> Optional[str]:
+    def pr_id(self) -> Optional[str]:
         return os.getenv("CI_MERGE_REQUEST_IID")
 
     @cachedproperty
-    def mr_title(self) -> Optional[str]:
+    def pr_title(self) -> Optional[str]:
         return os.getenv("CI_MERGE_REQUEST_TITLE")
 
     def to_dict(self) -> Dict[str, Any]:
```
brendongo: Should we remove the override since this is behavior of base class?

src/semgrep_agent/meta.py
```diff
@@ -35,13 +35,18 @@ def event_name(self) -> str:
...
 
     @cachedproperty
     def repo_name(self) -> str:
-        return Path.cwd().name
+        repo = gitpython.Repo(".", search_parent_directories=True)
+        return str(os.path.basename(repo.working_tree_dir))
+
+    @cachedproperty
+    def repo_url(self) -> Optional[str]:
+        return os.getenv("REPO_URL")
```
brendongo: For all the baseclass properties that read env variables do we expect notifications backend to still work without these fields? Can we comment somewhere which fields are necessary and which are optional. Would be useful for future classes that inherit from this to know which ones it needs to fill in and which ones are fine to leave null. Do we fail loudly if a required env variable is not set when we expect it?

src/semgrep_agent/meta.py
chmccreery: 🤷‍♀️ I feel like it is nice to be able to call from anywhere inside your repo

src/semgrep_agent/meta.py
chmccreery: I just grouped them in line 60 based on required/not. Only "repository" is required, and notifications can work without the rest (although they are nice to have to make the notification more useful). But yes, we now raise a RuntimeError in line 46 if this doesn't get set properly and no commandline arguments are _necessary_ for notifications to work

src/semgrep_agent/meta.py
chmccreery: yes, good catch

chmccreery: wait, are you talking about the to_dict method?

src/semgrep_agent/meta.py
underyx: We could get rid of some duplication by using `self.repo` here.

src/semgrep_agent/meta.py
underyx: Maybe we should prefix all these env variable names with `SEMGREP_` so it's `SEMGREP_REPO_URL`?

- Someone might have other software (e.g. CI) running that uses the same names
- It's nice when debugging an environment to know what each variable belongs to

src/semgrep_agent/meta.py
brendongo: Yeah? It looks like the to_dict is just a pass through to the parent class?

brendongo (approved):

src/semgrep_agent/meta.py
```diff
@@ -57,30 +63,46 @@ def commit(self) -> gitpython.Commit:  # type: ignore
...
             "config": self.config,
             "on": self.event_name,
-            "branch": None,
-            "pull_request_timestamp": None,
             "pull_request_author_username": None,
             "pull_request_author_image_url": None,
-            "pull_request_id": None,
-            "pull_request_title": None,
-            "semgrep_version": sh.semgrep(version=True).strip(),
-            "python_version": sh.python(version=True).strip(),
```
chmccreery, line 83: None of these are used by the semgrep_app backend, so I assumed it was ok to delete them. @underyx does that seem right to you or am I missing something?

src/semgrep_agent/meta.py
chmccreery: Ok, thanks for the clarification. Just fixed... GH interface shows the 4 lines above where you commented and it highlights the diffs instead of the line you commented on, so looked misleading.

src/semgrep_agent/meta.py
nbrahms: I thought we used the pull_request_id and pull_request_title.

nbrahms (approved):

PR #109, opened 2020-11-17
Add '--no-rewrite-rule-ids' to Semgrep runs to fix 'nosem' functionality

### LINKED ISSUE #2033: Inline // nosem not working
**Describe the bug**
When chef runs Semgrep in CI, they have several lines that get flagged every time, even though they are tagged with comments like // nosem: ruleid

**To Reproduce**
Create a file in go that violates the rule go.grpc.security.grpc-client-insecure-connection.grpc-client-insecure-connection and add a // nosem: go.grpc.security.grpc-client-insecure-connection.grpc-client-insecure-connection to it. Then run semgrep 

**Expected behavior**
This line of code should not show up in Semgrep findings on the commandline

**What is the priority of the bug to you?**
P2 - this was a P0 for msorens, but since // nosem without rule id specified seems to work, he is unblocked now. Others may still become blocked, too, though.

**Environment**
Docker using custom Buildkite setup
Language: Go

mschwager:
Fixes https://github.com/returntocorp/semgrep/issues/2033.

Before:

```go
func conn() grpc.ClientConnInterface {
	cc, err := grpc.Dial("localhost:1234", grpc.WithInsecure())
	if err != nil {
		panic(err)
	}
	return cc
}
```

```
$ python -m semgrep_agent --publish-deployment ... --publish-token ... 
...
go.grpc.security.grpc-client-insecure-connection.grpc-client-insecure-connection
     > test.go:2
     ╷
    2│   cc, err := grpc.Dial("localhost:1234", grpc.WithInsecure())
     ╵
     = Found an insecure gRPC connection. This creates a connection without
       encryption to a gRPC server. A malicious attacker could tamper with the
       gRPC message, which could compromise the machine.

=== exiting with failing status
```

```go
func conn() grpc.ClientConnInterface {
	cc, err := grpc.Dial("localhost:1234", grpc.WithInsecure()) // nosem: go.grpc.security.grpc-client-insecure-connection.grpc-client-insecure-connection
	if err != nil {
		panic(err)
	}
	return cc
}
```

```
$ python -m semgrep_agent --publish-deployment ... --publish-token ... 
...
| No current issues found
=== not looking at pre-existing issues since there are no current issues
=== exiting with success status
```

### DIFF
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -137,7 +137,12 @@ def invoke_semgrep(
             "=== looking for current issues in " + unit_len(paths, "file"), err=True
         )
         for chunk in chunked_iter(paths, PATHS_CHUNK_SIZE):
-            args = ["--skip-unknown-extensions", "--json", *config_args]
+            args = [
+                "--skip-unknown-extensions",
+                "--json",
+                "--no-rewrite-rule-ids",
+                *config_args,
+            ]
             for path in chunk:
                 args.append(path)
             semgrep_results = json.loads(str(semgrep(*args)))["results"]
@@ -165,7 +170,12 @@ def invoke_semgrep(
                     err=True,
                 )
                 for chunk in chunked_iter(paths_to_check, PATHS_CHUNK_SIZE):
-                    args = ["--skip-unknown-extensions", "--json", *config_args]
+                    args = [
+                        "--skip-unknown-extensions",
+                        "--json",
+                        "--no-rewrite-rule-ids",
+                        *config_args,
+                    ]
                     for path in chunk:
                         args.append(path)
                     semgrep_results = json.loads(str(semgrep(*args)))["results"]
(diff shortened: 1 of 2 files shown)


chmccreery (approved):
thank you for fixing this 🙏

chmccreery:
It looks from the images above like you tested that rules without the // nosem are now printed out correctly too (without the tmp. but with the rest of the prefixes) but just want to make sure I interpreted that correctly

src/semgrep_agent/semgrep.py
```diff
@@ -128,21 +128,23 @@ def invoke_semgrep(
...
             "=== looking for current issues in " + unit_len(paths, "file"), err=True
         )
         for chunk in chunked_iter(paths, PATHS_CHUNK_SIZE):
-            args = ["--skip-unknown-extensions", "--json", *config_args]
+            args = [
+                "--skip-unknown-extensions",
+                "--json",
+                "--no-rewrite-rule-ids",
+                *config_args,
+            ]
```
chmccreery, line 143: if uses_managed policy else (do not include the --no-rewrite-rule-ids) per Bence's commment on slack

chmccreery (changes requested):

chmccreery (approved):

PR #110, opened 2020-11-17
Add environment variables to displayed commandline output

chmccreery:
When debugging, it is useful to know which environment variables have been properly detected by the system.

Policy printing will work once https://github.com/returntocorp/semgrep-app/pull/1126 is deployed

Closes https://github.com/returntocorp/semgrep-app/issues/1107

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -88,11 +88,12 @@ def main(
     # Setup URL/Token
     sapp = Sapp(url=publish_url, token=publish_token, deployment_id=publish_deployment)
     maybe_print_debug_info(meta)
-    sapp.report_start(meta)
+    policy = sapp.report_start(meta)
     if sapp.is_configured:
         click.echo(
             f"| {server} - logged in as deployment #{sapp.deployment_id}", err=True,
         )
+        click.echo(f"| using policy {policy}")
     else:
         click.echo(f"| {server} - not logged in", err=True)
 
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -72,7 +72,7 @@ def main(
     )
 
     # Get Metadata
-    server = os.getenv("INPUT_PUBLISHURL") or "https://semgrep.dev"
+    # server = os.getenv("INPUT_PUBLISHURL") or "https://semgrep.dev"
     Meta = detect_meta_environment()
     meta_kwargs = {}
     if baseline_ref:
@@ -90,15 +90,22 @@ def main(
     maybe_print_debug_info(meta)
     policy = sapp.report_start(meta)
     if sapp.is_configured:
-        click.echo(
-            f"| {server} - logged in as deployment #{sapp.deployment_id}", err=True,
-        )
+        if publish_url == "https://semgrep.dev":
+            click.echo(
+                f"| Semgrep Community - logged in as deployment #{sapp.deployment_id}",
+                err=True,
+            )
+        else:
+            click.echo(
+                f"| Semgrep Community - logged in to {publish_url} as deployment #{sapp.deployment_id}",
+                err=True,
+            )
         if policy:
             click.echo(f"| policy - using {policy}")
         else:
             click.echo(f"| policy - unknown")
     else:
-        click.echo(f"| {server} - not logged in", err=True)
+        click.echo(f"| Semgrep Community - not logged in", err=True)
 
     for env_var in [
         "SEMGREP_REPO_URL",
@@ -108,7 +115,7 @@ def main(
         "SEMGREP_PR_TITLE",
     ]:
         if os.getenv(env_var):
-            click.echo(f"| {env_var}={os.getenv(env_var)}")
+            click.echo(f"| {env_var} - {os.getenv(env_var)}")
 
     # Setup Config
     click.echo("=== setting up agent configuration", err=True)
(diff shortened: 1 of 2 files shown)


chmccreery:
Screen Shot 2020-11-17 at 12 30 45 PM

chmccreery:
Screen Shot 2020-11-17 at 4 49 05 PM

src/semgrep_agent/semgrep_app.py
```diff
@@ -74,6 +74,7 @@ def report_start(self, meta: GitMeta) -> None:
                 ignore_patterns=glom(body, T["scan"]["meta"].get("ignored_files", [])),
             )
             debug_echo(f"=== Our scan object is: {self.scan!r}")
+            return str(glom(body, T["policy"])) if "policy" in body else None
```
underyx:
```suggestion
            return glom(body, T["policy"], default=None)
```

Could you try if this is equivalent? I'm 90% sure yes.

src/semgrep_agent/main.py
```diff
@@ -87,13 +88,25 @@ def main(
...
     maybe_print_debug_info(meta)
-    sapp.report_start(meta)
+    policy = sapp.report_start(meta)
     if sapp.is_configured:
         click.echo(
-            f"| semgrep.dev - logged in as deployment #{sapp.deployment_id}", err=True,
+            f"| {server} - logged in as deployment #{sapp.deployment_id}", err=True,
         )
+        if policy:
+            click.echo(f"| using policy {policy}")
```
```suggestion
            click.echo(f"| policy - using {policy}")
        else:
            click.echo(f"| policy - unknown")
```

src/semgrep_agent/main.py
```diff
@@ -87,13 +88,25 @@ def main(
...
+
+    for env_var in [
+        "SEMGREP_REPO_URL",
+        "SEMGREP_JOB_URL",
+        "SEMGREP_BRANCH",
+        "SEMGREP_PR_ID",
+        "SEMGREP_PR_TITLE",
+    ]:
+        if os.getenv(env_var):
+            click.echo(f"| {env_var}={os.getenv(env_var)}")
```
```suggestion
            click.echo(f"| {env_var} - {os.getenv(env_var)}")
```

Just keeping the same format as existing lines.

src/semgrep_agent/main.py
```diff
@@ -72,6 +72,7 @@ def main(
     )
 
     # Get Metadata
+    server = os.getenv("INPUT_PUBLISHURL") or "https://semgrep.dev"
```
underyx: The click library gives us this value in the `publish_url` function parameter. That one is actually better, cause env vars are not the only way to override the URL, you could also use `--publish-url` on the command line.

So I'd just use the `publish_url` variable later instead of setting `server`.

src/semgrep_agent/main.py
```diff
@@ -87,13 +88,25 @@ def main(
     # Setup URL/Token
     sapp = Sapp(url=publish_url, token=publish_token, deployment_id=publish_deployment)
     maybe_print_debug_info(meta)
-    sapp.report_start(meta)
+    policy = sapp.report_start(meta)
     if sapp.is_configured:
         click.echo(
-            f"| semgrep.dev - logged in as deployment #{sapp.deployment_id}", err=True,
+            f"| {server} - logged in as deployment #{sapp.deployment_id}", err=True,
```
```suggestion
            f"| Semgrep Community - logged in as deployment #{sapp.deployment_id}", err=True,
```

I'd conditionally add the URL, only if it's not the default. Something like:

```
logged in to dev.semgrep.dev as deployment #1
```

underyx (changes requested):

src/semgrep_agent/semgrep_app.py
chmccreery: mypy gets upset if you don't cast the return as either a None or a string, and if you add str(...) around the whole thing here it will not return None properly. Other ideas? I could just make return type Any but I think that is just trading one form of clunkiness for another

src/semgrep_agent/main.py
chmccreery: I don't feel _that_ strongly about this, but I do think it looks wrong in the case of environment variables to use a dash instead of equals. I prefer the consistency with how they are set in the terminal.

src/semgrep_agent/semgrep_app.py
underyx:
```suggestion
            from typing import cast, Optional
            return cast(Optional[str], glom(body, T["policy"], default=None))
```

I just found this mypy feature an hour ago, let's try it out? :D

src/semgrep_agent/main.py
underyx: Fair enough!

chmccreery:
Screen Shot 2020-11-19 at 12 42 20 PM

src/semgrep_agent/semgrep_app.py
```diff
@@ -74,7 +75,7 @@ def report_start(self, meta: GitMeta) -> Optional[str]:
                 ignore_patterns=glom(body, T["scan"]["meta"].get("ignored_files", [])),
             )
             debug_echo(f"=== Our scan object is: {self.scan!r}")
-            return str(glom(body, T["policy"])) if "policy" in body else None
+            return cast(Optional[str], glom(body, T["policy"], default=None))
```
underyx, line 78: hahaha it's exactly the same amount of characters in the end

but we did it :D

underyx (approved):
Looking good!

PR #111, opened 2020-11-18
Send list of ignored findings to semgrep.dev in managed mode

underyx:


chmccreery:
Lgtm! Is this one of those things where it will only pass live runs test after some changes to semgrep and semgrep-app are deployed?

chmccreery (approved):

underyx:
> Lgtm! Is this one of those things where it will only pass live runs test after some changes to semgrep and semgrep-app are deployed?

Correct, it's waiting for a semgrep.dev deploy!

PR #116, opened 2020-11-19
Fix crash when a newly created file yields issues

underyx:
The target manager used to yield all paths as paths to check on the
baseline commit. This used to be okay since when these are passed to
semgrep with --include, they are silently ignored. But since we switched
to explicit targeting, passing non-existent paths causes a crash.

### DIFF
--- a/src/semgrep_agent/targets.py
+++ b/src/semgrep_agent/targets.py
@@ -278,9 +278,12 @@ class TargetFileManager:
         """
         Prepare file system for baseline scan, and return the paths to be analyzed.
 
-        Returned list of paths are all abolute paths and include all files that are
+        Returned list of paths are all relative paths and include all files that are
+            - already in the baseline commit, i.e. not created later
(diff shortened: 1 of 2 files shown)


src/semgrep_agent/targets.py
```diff
@@ -285,11 +285,18 @@ def baseline_paths(self) -> Iterator[List[Path]]:
         :return: A list of paths
         :raises ActionFailure: If git cannot detect a HEAD commit or unmerged files exist
         """
-        if self._base_commit is None:
+        repo = get_git_repo()
```
brendongo, line 291: Probably worth mentioning in the function comment that this returns empty list if not a git repo and otherwise returns elements in target paths that were not in added

brendongo (approved):

src/semgrep_agent/targets.py
underyx: good call! added

PR #119, opened 2020-11-19
Adds gitlab output support via --gitlab flag

chgg-kboberg:
Attn: @underyx 

Adds gitlab-formatted output for the Gitlab action.  Should (if I understand _Click behavior correctly) automagically set gitlab output if run in a Gitlab environment.

Also comes with some basic bootstrapping of test infrastructure!


pyproject.toml
```diff
@@ -18,6 +18,9 @@ gitpython = "^2.1.15"
 glom = "^20.8.0"
 pymmh3 = "^0.0.5"
 attrs = "~=19.3.0"
+pytest = "^6.1.2"
+pathlib = "^1.0.1"
+jsonschema = "^3.2.0"
```
underyx: could you install these as dev dependencies instead (`poetry add --dev`)? also, pathlib is already in the standard library of the Python version we use, so you might be able to skip it

src/semgrep_agent/findings.py
```diff
@@ -98,6 +99,61 @@ def to_dict(self, omit: Set[str]) -> Mapping[str, Any]:
...
+        return {
+            "id": str(uuid.uuid5(uuid.NAMESPACE_URL, str(self.path))),
+            "category": "sast",
+            # CVE is a required field from Gitlab schema.  Semgrep is CVE-unaware AFAIK
+            "cve": "",
+            "message": self.message,
+            "severity": self._to_gitlab_severity(),
+            # KB note:
+            # Semgrep is designed to be a low-FP tool by design.
+            # Does hard-coding confidence make sense here?
```
underyx, line 112: yep, we did this for SARIF as well

although now I think we should've done `"High" if "audit" not in rule_id.split(".") else "Low"` (the audit rules are to be used with human reviewers checking all results)

src/semgrep_agent/main.py
```diff
@@ -142,6 +146,14 @@ def main(
...
         ]
         click.echo(json.dumps(output))
+    elif gitlab_output:
+        #output all new findings in Gitlab format
+        output = {
+            "$schema": "https://gitlab.com/gitlab-org/security-products/security-report-schemas/-/blob/master/dist/sast-report-format.json",
+            "version": "2.0",
+            "vulnerabilities": [f.to_gitlab() for f in new_findings],
+        }
+        click.echo(json.dumps(output))
```
underyx: We could even write it directly to a file maybe, if that's what GitLab needs and skip using a pipe. Or do they understand results from stdout as well?

pytest/conftest.py
```diff
@@ -0,0 +1,94 @@
...
+    if output_format == "gitlab":
+        options.append("--gitlab")
+
+    process = subprocess.Popen(
+        ["python3", "-m", "semgrep_agent", *options],
+        encoding="utf-8",
+        stderr=subprocess.STDOUT if stderr else subprocess.PIPE,
+        stdout=subprocess.PIPE
+    )
+    # we have to use Popen,wait,communicate here because findings trigger non-zero exit codes
```
underyx: I think `subprocess.run` would have a nicer API that doesn't raise on non-zero exit codes.

src/semgrep_agent/main.py
chgg-kboberg: File support is an option.  Gitlab artifact definitions look like the following:

```yaml
goose-sast:
  extends: .sast-analyzer
  image: kboberg/goose:latest
  tags:
    - security
  artifacts:
    reports:
      sast: gl-sast-report.json
  script:
    - /app/goose -target $CI_PROJECT_DIR -config /app/config/goose_rules.json -decisiontree -outputmode gitlab > gl-sast-report.json
````

Comment (underyx) [reply src/semgrep_agent/main.py]: cool, let's leave it this way then if other tools tend to use `cmd > gl-sast-report.json` syntax as well 👍

File: pytest/conftest.py (hunk 4 above)
Comment (underyx) [reply pytest/conftest.py]: Even better, we could:

- pass a `cwd` parameter there with the directory that has that `stupid.py` with known semgrep issues
- and we can rename the `pytest/` directory back to a more standard `tests/` (it makes me uncomfortable, I worry somehow `import pytest` will think this is the correct module directory 😅 )

File: CONTRIBUTING.md [hunk 5]
```diff
@@ -14,6 +14,19 @@ Install dependencies with poetry
...
+```
+poetry shell
+```
+
+Install pre-commit hooks:
+
+```
+python -m pip install pre-commit
+pre-commit install
+```
```
underyx, line 27: 😻 thanks so much!

underyx:
"semgrep with managed policy" only fails cause your fork doesn't have access to our API token by the way, don't worry about that job

chgg-kboberg:
Tests now live in the `tests/` directory once more!

image

underyx:
Awesome job, can't thank you enough! Especially for the testing infrastructure ☺️

PR #120, opened 2020-11-19
allow manual override of repo name

chmccreery:
using optional environment variable: SEMGREP_REPO_NAME
adds vertical alignment to printed statements

fixes https://github.com/returntocorp/semgrep-app/issues/1105

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -36,8 +36,11 @@ class CliObj:
 def get_event_type() -> str:
     if "GITHUB_ACTIONS" in os.environ:
         return os.environ["GITHUB_EVENT_NAME"]
-
     return "push"
+
+
+def get_aligned_command(title: str, subtext: str) -> str:
+    return f"| {title.ljust(17)} - {subtext}"
 
 
 @click.command()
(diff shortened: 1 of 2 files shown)


src/semgrep_agent/main.py
```diff
@@ -36,10 +36,13 @@ class CliObj:
 def get_event_type() -> str:
     if "GITHUB_ACTIONS" in os.environ:
         return os.environ["GITHUB_EVENT_NAME"]
-
     return "push"
 
 
+def get_aligned_command(title: str, subtext: str) -> str:
+    return f"| {title.ljust(17)} - {subtext}"
```
underyx, line 43: 😻

src/semgrep_agent/main.py
```diff
@@ -90,32 +94,28 @@ def main(
...
-        click.echo(f"| Semgrep Community - not logged in", err=True)
+        click.echo(get_aligned_command("manage", f"not logged in"), err=True)
 
     for env_var in [
         "SEMGREP_REPO_URL",
         "SEMGREP_JOB_URL",
         "SEMGREP_BRANCH",
         "SEMGREP_PR_ID",
         "SEMGREP_PR_TITLE",
+        "SEMGREP_REPO_NAME",
```
underyx: They get printed in this order as well, so could we maybe move this up to be next to REPO_URL?

src/semgrep_agent/meta.py
```diff
@@ -43,7 +43,9 @@ def repo(self) -> gitpython.Repo:  # type: ignore
     def repo_name(self) -> str:
         if not self.repo.head.is_valid():
             raise RuntimeError("Semgrep action cannot run on repository with no HEAD")
-        return str(os.path.basename(self.repo.working_tree_dir))
+        return os.getenv("SEMGREP_REPO_NAME") or str(
```
underyx, line 46: If the repo name is set from an env var, we probably don't need to raise an error above, we could move the first possible return to above that if condition.

underyx (approved):
Two non-blocking nitpicks, but looks good!

src/semgrep_agent/meta.py
chmccreery: I feel like this actually goes back to a question of do we require semgrep-action to be run on a git-controlled repo. If not, I think we can do as you say, but otherwise I still think this check is important because without a valid repo object, other things in the codebase may break. In the case of buildkite users, it seems like we still want to check for a valid repo, we just also want to let them customize name to include org_name, for example

PR #121, opened 2020-11-20
.github: Add pytest workflow

underyx:


.github/workflows/test.yml
```diff
@@ -3,11 +3,55 @@ name: Tests
...
+        with:
+          path: ${{ steps.setup-poetry.outputs.cache-dir }}
+          key: ${{ runner.os }}-${{ matrix.python-version }}-poetry-${{ hashFiles('**/poetry.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ matrix.python-version }}-poetry-
+
+      - name: Install dependencies
+        run: |
+          poetry install
+          poetry add semgrep --no-interaction
```
underyx: 1. the unpinned version bothers me
2. not sure if installing with poetry or with pipx (or pip) is better

tests/conftest.py
```diff
@@ -9,10 +9,7 @@
 
 # Large swaths of this test infrastructure is shamelessly stolen from semgrep core test infrastructure
 TESTS_PATH = Path(__file__).parent
-MASKED_KEYS = [
-    "tool.driver.semanticVersion",
-    "results.extra.metavars.*.unique_id.md5sum",
-]
+MASKED_KEYS: List[str] = []
```
nbrahms: I don't understand this change...

If we're removing the functionality, shouldn't we just delete the constant and its use?

.github/workflows/test.yml
brendongo: Why not just pin it? We update semgrep version manually anyway. Just add the it should also be bumped here?

tests/conftest.py
```diff
@@ -39,8 +36,13 @@ def _mark_masked(obj, path_items):
...
 
 def _clean_output_json(output_json: str) -> str:
-    """Make semgrep's output deterministic and nicer to read."""
-    output = json.loads(output_json)
+    """Make JSON output deterministic and nicer to read."""
+    try:
+        output = json.loads(output_json)
+    except json.JSONDecodeError:
+        raise ValueError(
+            f"Instead of JSON, output was:\n--- output start ---\n{output_json}\n--- output end ---"
```
nbrahms, line 20:
```suggestion
            f"Instead of JSON, Semgrep output was:\n--- output start ---\n{output_json}\n--- output end ---"
```

brendongo (approved):

tests/conftest.py
nbrahms: probably also use printing the semgrep stderr in this case somehow?

tests/conftest.py
```diff
@@ -70,13 +70,15 @@ def _run_semgrep_agent(
         options.append("--gitlab-json")
 
     process = subprocess.run(
-        ["python3", "-m", "semgrep_agent", *options],
+        ["semgrep-agent", *options],
```
nbrahms: I think this does not work in some environments...

.github/workflows/test.yml
nbrahms: I think you'd want to update the bump script in that case, as well

tests/conftest.py
underyx: it's printed above and pytest shows captured print output when a test fails

.github/workflows/test.yml
underyx: No bump script here unfortunately, and wanted to avoid adding more manual work :(

tests/conftest.py
underyx: Yeah, I wanted to add snapshot tests soon and wanted to also make it obvious that the existing elements were not used.

tests/conftest.py
chmccreery: can verify ^

underyx:
@brendongo can you re-review the part where we install semgrep for the pytest job?

.github/workflows/test.yml
```diff
@@ -3,11 +3,72 @@ name: Tests
...
+        if: steps.reuse-poetry-virtualenv.outputs.cache-hit != 'true'
+
+      - name: Install semgrep-agent
+        run: poetry install
+
+      - name: Run pytest
+        run: |
+          # tests should simulate CI environment iff they need one
+          unset CI
+          unset "${!GITHUB_@}"
```
brendongo, line 66: Does this unset all GITHUB_ env variables?

brendongo (approved):

.github/workflows/test.yml
underyx: yep

PR #122, opened 2020-11-20
Allow untracked paths to be present

underyx:
But we still need to worry about the edge case where someone removes a path since the baseline, and now has the same path as untracked. That should not be allowed.

Changes were tested manually with the following commands:

    git rm --cached README.md
    echo "5 == 5" > bad.py
    git add bad.py
    git commit -m "preparation commit"
    poetry run python -m semgrep_agent --config p/python --baseline-ref HEAD~1
    // should see a complaint that README changed and is untracked

    rm README.md
    touch new-file
    poetry run python -m semgrep_agent --config p/python --baseline-ref HEAD~1
    // should see successful scan with 5 == 5

    echo "hiii" > LICENSE
    poetry run python -m semgrep_agent --config p/python --baseline-ref HEAD~1
    // should see a complaint that there are changes to tracked files

    git add LICENSE
    poetry run python -m semgrep_agent --config p/python --baseline-ref HEAD~1
    // should see a complaint that there are changes to tracked files

### DIFF
--- a/src/semgrep_agent/targets.py
+++ b/src/semgrep_agent/targets.py
@@ -3,6 +3,7 @@ from contextlib import contextmanager
 from enum import auto
 from enum import Enum
 from pathlib import Path
+from typing import Dict
 from typing import Iterator
 from typing import List
 from typing import NamedTuple
@@ -13,6 +14,7 @@ from typing import TYPE_CHECKING
 import attr
 import click
 import sh
+from boltons.iterutils import bucketize
 from boltons.strutils import unit_len
 from sh.contrib import git
 
@@ -42,8 +44,9 @@ class StatusCode:
     Renamed = "R"
     Modified = "M"
     Unmerged = "U"
-    Untracked = "?"
     Ignored = "!"
+    Untracked = "?"
+    Unstaged = " "  # but changed
 
 
 @attr.s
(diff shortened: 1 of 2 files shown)


nbrahms:
How did you test success?

src/semgrep_agent/targets.py
```diff
@@ -208,17 +214,58 @@ def _get_target_files(self) -> List[Path]:
 
         return relative_paths
 
-    def _abort_if_dirty(self) -> None:
+    @_dirty_paths_by_status.default
+    def get_dirty_paths_by_status(self) -> Dict[str, List[Path]]:
```
nbrahms, line 218:
```suggestion
    def get_dirty_paths_by_status(self) -> Mapping[str, Sequence[Path]]:
```

underyx:
> How did you test success?

See PR description / commit message. The lines marked with `//` are my manual assertions.

nbrahms (approved):

src/semgrep_agent/targets.py
underyx: @nbrahms aren't the more specific ones better? With the earlier annotation, dict methods would be suggested on the attribute as well.

PR #125, opened 2020-12-02
Send empty findings lists to semgrep.dev as well

underyx:
Send empty findings lists to semgrep.dev as well.

Snuck a bunch of dependency bumps in here as well; tested the gitpython bump by running with `--baseline-ref 'HEAD~30'` locally successfully. Their major version bump was only because they dropped python 2 support.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,19 +1,18 @@
 [[package]]
-category = "dev"
-description = "Atomic file writes."
-marker = "sys_platform == \"win32\""
 name = "atomicwrites"
+version = "1.4.0"
+description = "Atomic file writes."
+category = "dev"
 optional = false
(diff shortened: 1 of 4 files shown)


src/semgrep_agent/constants.py
```diff
@@ -1,3 +1,5 @@
 SUPPORT_EMAIL = "support@r2c.dev"
 
 PRIVACY_SENSITIVE_FIELDS = {"syntactic_context"}
+
+FINDINGS_UPLOAD_CHUNK_SIZE = 10_000
```
nbrahms: looks like you can just delete this

nbrahms (approved):
Nice catch

PR #128, opened 2020-12-03
Add explicit check for publish token length

mschwager:


src/semgrep_agent/constants.py
```diff
@@ -1,3 +1,5 @@
 SUPPORT_EMAIL = "support@r2c.dev"
 
 PRIVACY_SENSITIVE_FIELDS = {"syntactic_context"}
+
+PUBLISH_TOKEN_LENGTH = 16
```
mschwager: After generating a test token in the cloud app UI, it looks like the length is 64, not 16. Just wanted to double check.

nbrahms: ah whoops

nbrahms:
```postgres=> select distinct char_length(token) from token;
 char_length
-------------
          60
(1 row)```

File: src/semgrep_agent/semgrep_app.py [hunk 2]
```diff
@@ -43,6 +43,11 @@ def __post_init__(self) -> None:
         #
         if self.token and self.deployment_id:
             self.is_configured = True
+        if self.is_configured and len(self.token) < constants.PUBLISH_TOKEN_LENGTH:
```
nbrahms: I'd probably make a regex:

```
TOKEN_FORM = re.compile("\w{64,}")

...

if TOKEN_FORM.match(self.token) ...:
```

to allow us to increase its length later.

src/semgrep_agent/constants.py
mschwager:
```suggestion
PUBLISH_TOKEN_LENGTH = 64
```

underyx (approved):
This looks good to me. Just to take notes for the future, possible upgrade paths are:

- Doing validation from the click CLI option definition
- Returning a clear error message like here from semgrep.dev instead, and just displaying that message in the agent

The second one will be especially nice cause 1) we'll have central records of when/how often token failures happened and 2) the expected token format will be easier to keep in sync.

mschwager:
> Doing validation from the click CLI option definition

Yup, that seems like a better place to perform this check. Although, we may want to check in both locations. The current check will validate any time a `Sapp` object is instantiated, whereas the `click` check will provide a better UX.

PR #132, opened 2020-12-07
Ensure GitHub performs baseline against PR head, not merge commit

### LINKED ISSUE #131: Fix PR merge result being compared to base branch in GHA
the issue is with a surprising behavior of github: let’s assume you have this git history:

```
main branch  0--1--2--3--4--5
                 \
your branch       A--B
```

when github starts this job, it actually merges commit B into 5! that’s the codebase semgrep sees and our diff-aware scanning will compare “B merged into 5” against 1 to find new issues

so while you expect that only A and B’s changes will be scanned, right now it’s actually A, B, and 2 through 5 all being scanned, hence the additional changed file count

This might be the right one to use? But we still probably need to add python logic to fetch more commits for the baseline checkout to work https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit

nbrahms:
Also adds some helpful logging to show exactly what's being scanned.

Fixes #131 

Best reviewed with whitespace changes hidden.

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -150,12 +150,9 @@ def main(
         click.echo(message, err=True)
         sys.exit(1)
 
-    committed_datetime = meta.commit.committed_datetime if meta.commit else None
-
     results = semgrep.scan(
         config,
-        committed_datetime,
-        meta.base_commit_ref,
+        meta,
         semgrep.get_semgrepignore(sapp.scan.ignore_patterns),
         sapp.is_configured,
     )
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -150,9 +150,13 @@ def main(
         click.echo(message, err=True)
         sys.exit(1)
 
+    committed_datetime = meta.commit.committed_datetime if meta.commit else None
+
     results = semgrep.scan(
         config,
-        meta,
+        committed_datetime,
+        meta.base_commit_ref,
+        meta.head_ref,
         semgrep.get_semgrepignore(sapp.scan.ignore_patterns),
         sapp.is_configured,
     )
--- a/src/semgrep_agent/meta.py
+++ b/src/semgrep_agent/meta.py
@@ -5,6 +5,7 @@ from dataclasses import dataclass
 from dataclasses import field
 from pathlib import Path
 from typing import Any
+from typing import cast
 from typing import Dict
 from typing import Optional
 from typing import Type
@@ -54,6 +55,10 @@ class GitMeta:
     @cachedproperty
     def commit_sha(self) -> Optional[str]:
         return self.repo.head.commit.hexsha  # type: ignore
+
+    @cachedproperty
+    def head_ref(self) -> Optional[str]:
+        return "HEAD"
 
     @cachedproperty
     def base_commit_ref(self) -> Optional[str]:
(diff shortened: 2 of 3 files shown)


nbrahms:
Tested by manually setting `head_ref` in code.

src/semgrep_agent/main.py
```diff
@@ -150,12 +150,9 @@ def main(
         click.echo(message, err=True)
         sys.exit(1)
 
-    committed_datetime = meta.commit.committed_datetime if meta.commit else None
-
     results = semgrep.scan(
         config,
-        committed_datetime,
-        meta.base_commit_ref,
+        meta,
```
brendongo: Can we explicitly pass down head_ref. I'd rather not mix the meta spaghetti with the semgrep spaghetti. IMO much easier for us to rewrite later when the information dependency is explicit even with the pain of having to pass variables in many places.

nbrahms:
Additional testing with various dirty states:

staged+unstaged modifications / deletions / adds on "push" mode.

src/semgrep_agent/meta.py
```diff
@@ -145,12 +150,18 @@ def commit_sha(self) -> Optional[str]:
         return super().commit_sha  # type: ignore
 
     @cachedproperty
-    def base_commit_ref(self) -> Optional[str]:
+    def head_ref(self) -> Optional[str]:
```
brendongo, line 153: What do we expect this to return/point to.

src/semgrep_agent/semgrep.py
```diff
@@ -107,61 +110,112 @@ def rewrite_sarif_file(sarif_path: Path) -> None:
...
+    if get_git_repo() is None:
+        yield base_ref
+        return
+
+    if base_ref:
+        # Preserve location of head^ after we possibly change location below
+        base_ref = git(["rev-parse", base_ref]).stdout.decode("utf-8").rstrip()
+
+    if meta.head_ref:
+        stashed_rev = git(["branch", "--show-current"]).stdout.decode("utf-8").rstrip()
```
brendongo, line 139: `git branch --show-current` doesn't work for me. Is this relying on a git feature people might not have?

src/semgrep_agent/meta.py
```diff
@@ -55,6 +56,10 @@ def repo_url(self) -> Optional[str]:
     def commit_sha(self) -> Optional[str]:
         return self.repo.head.commit.hexsha  # type: ignore
 
+    @cachedproperty
+    def head_ref(self) -> Optional[str]:
```
brendongo, line 60: Is it fine that this is set to None for non Github projects?

brendongo (changes requested):
So for a PR:

B - 1 - 2 - 3 
 \ 
 X - Y - Z

If head_ref points to Z what I think this PR is doing is comparing findings from Z to findings from B but don't we want to compare findings from Z to findings on 3?

nbrahms:
> So for a PR:
> 
> B - 1 - 2 - 3
> \
> X - Y - Z
> 
> If head_ref points to Z what I think this PR is doing is comparing findings from Z to findings from B but don't we want to compare findings from Z to findings on 3?

In your example we'd want to compare Z to B (not to 3). This is because you don't expect to see results from other devs' PRs.

src/semgrep_agent/semgrep.py
nbrahms: Since this only applies in GHA, I think it is not; since it works in our GHA environment it should be sufficient (also, wow, I think you need to update your git!)

src/semgrep_agent/meta.py
nbrahms: Yes, if this is `None`, the entire work here is a no-op.

brendongo:
When testing on a PR I would expect it to show me findings that would be introduced if the PR were to be merged. I think if we compare Z to B we potentially miss findings that would be introduced if Z is merged onto 3.

src/semgrep_agent/meta.py
nbrahms: It should point to the PR branch head ref.

nbrahms:
> When testing on a PR I would expect it to show me findings that would be introduced if the PR were to be merged. I think if we compare Z to B we potentially miss findings that would be introduced if Z is merged onto 3.

That is specifically not the behavior expected by our users, however, and has been leading to a bunch of confusion re. duplicate results (as all PRs with a base below an introduced finding will trigger).

brendongo:
Clarified on video call:

The problem users had was seeing findings from 1-2-3 in their PR for Z.

What this PR is doing is that it is avoiding 1-2-3 by comparing Z to B instead of comparing the merge commit of 3 and Z to B.

I can imagine a semgrep finding that does not exist in B-X-Y-Z but does exist in B-1-2-3-X-Y-Z which this change will miss. We are punting on this edge case for now but ideal solution is to figure out how to compare the merge commit of 3 and Z to 3.

brendongo (approved):

PR #134, opened 2020-12-08
fix merge refs

nbrahms:
Restores code from #131 , but also adds correct head ref sha from the GitHub event (was base ref, not head ref).

To test this, I'm going to make a one-off docker build tag and use that to test with a few customers and internally.

Diffs are best viewed with whitespace ignored.


nbrahms:
Tested internally using `returntocorp/semgrep-action@test`

brendongo (approved):

nbrahms:
Also tested with external users

PR #138, opened 2020-12-16
print number of rules being run if using web ui

### LINKED ISSUE #130: Semgrep action indicates how many rules ran on a scan

chmccreery:
- Gives a user confidence in which rules (and how many rules) ran on a given scan 
- Since policy name is printed here as well, it is easy to go to the semgrep-app for details
- helps to alert user that something is wrong (num_rules is 0 or is way larger than expected)

closes https://github.com/returntocorp/semgrep-action/issues/130
Screen Shot 2020-12-15 at 4 46 32 PM

### DIFF
--- a/src/semgrep_agent/semgrep_app.py
+++ b/src/semgrep_agent/semgrep_app.py
@@ -6,10 +6,12 @@ from pathlib import Path
 from typing import cast
 from typing import List
 from typing import Optional
+from typing import Tuple
 
 import requests
 from glom import glom
 from glom import T
+from ruamel import yaml  # type: ignore
 
 from semgrep_agent import constants
 from semgrep_agent.meta import GitMeta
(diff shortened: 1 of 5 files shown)


src/semgrep_agent/semgrep_app.py
```diff
@@ -104,13 +106,15 @@ def fetch_rules_text(self) -> str:
...
-    def download_rules(self) -> Path:
+    def download_rules(self) -> Tuple[Path, int]:
         """Save the rules configured on semgrep app to a temporary file"""
         # hey, it's just a tiny YAML file in CI, we'll survive without cleanup
         rules_file = tempfile.NamedTemporaryFile(suffix=".yml", delete=False)  # nosem
         rules_path = Path(rules_file.name)
-        rules_path.write_text(self.fetch_rules_text())
-        return rules_path
+        rules = self.fetch_rules_text()
+        parsed = yaml.load(rules, Loader=yaml.Loader)
```
underyx: I believe we use a different loader on the backend, the RT (or round-trip) loader. Is this intentionally different?

chmccreery: hmm I didn't see that and am still not finding it. Do you have a file name and line number I can look at? 

I tried using `yaml.load(...)` with no loader first, but it told me that was insecure and that I should use `Loader=yaml.Loader` unless that did not work for my purposes 🤷‍♀️

chmccreery: Oh, ok, just read https://yaml.readthedocs.io/en/latest/api.html and found line 36 in util.py in backend code. I'll update this

underyx (approved):

PR #147, opened 2021-01-06
Allow passing any arg with --semgrep-opts

underyx:
Tested by running:

```
python -m semgrep_agent --config p/r2c --baseline-ref HEAD~3 --semgrep-opts '--max-lines-per-finding 45 -j "2" --timeout 90'
```

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -61,6 +61,7 @@ def get_aligned_command(title: str, subtext: str) -> str:
 @click.option(
     "--gitlab-json", "gitlab_output", envvar="SEMGREP_GITLAB_JSON", is_flag=True
 )
+@click.option("--semgrep-opts", type=str)
 def main(
     config: str,
     baseline_ref: str,
@@ -69,6 +70,7 @@ def main(
     publish_deployment: int,
     json_output: bool,
     gitlab_output: bool,
+    semgrep_opts: str,
 ) -> NoReturn:
     click.echo("=== detecting environment", err=True)
     click.echo(
@@ -175,6 +177,7 @@ def main(
         meta.head_ref,
         semgrep.get_semgrepignore(sapp.scan.ignore_patterns),
         sapp.is_configured,
+        semgrep_opts=semgrep_opts,
     )
     new_findings = results.findings.new
 
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -57,11 +57,11 @@ def get_aligned_command(title: str, subtext: str) -> str:
 )
 @click.option("--publish-token", envvar="INPUT_PUBLISHTOKEN", type=str)
 @click.option("--publish-deployment", envvar="INPUT_PUBLISHDEPLOYMENT", type=int)
-@click.option("--json", "json_output", hidden=True, is_flag=True)
 @click.option(
     "--gitlab-json", "gitlab_output", envvar="SEMGREP_GITLAB_JSON", is_flag=True
 )
-@click.option("--semgrep-opts", type=str)
+@click.option("--json", "json_output", hidden=True, is_flag=True)
+@click.option("--semgrep-opts", hidden=True, type=str)
 def main(
     config: str,
     baseline_ref: str,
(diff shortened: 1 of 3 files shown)


minusworld:
Looks cool!

What does this prepare us for?  
Also, is there some possibility of injecting things via this configuration option that might be precarious from a security perspective? Things like injecting `--dangerously-allow...` or as an SSRF using config URLs. I'm uncertain of the architecture here myself so I wanted to ask if you had any insight there.

nbrahms:
I'd be much more comfortable with this if the options were validated using an allowlist. That way we don't have to worry about (e.g.) options that would break parsing, and allows us to more flexibly alter the API in the future.

underyx:
### Use cases & security notes

@minusworld the use case currently is to allow adjusting `--max-lines-per-finding` (to see more output when running git-aware semgrep locally), and to allow injecting a local rule to execute along with those defined in the policy. The security notes are great, I had not considered that. However, if someone has permission to change this bit of the CI config, they can also change 1) the semgrep-agent options, which already allow setting an arbitrary config URL and 2) the commands that run in CI, which is already arbitrary code execution in the CI context anyway.

### Why I'm not worried about no allowlist

@nbrahms I'm still in favor of the current implementation. This is meant to be an advanced flag, just like `--json`, which isn't even documented on the 'beaten path' docs. `--json` would also, in some way, break the agent (removes human-readable output), but we haven't seen a problem like that, and wouldn't expect them to crop up either.

I am also less worried about the risk of user-configured opts causing failures, because 1) if someone is messing with advanced options like this, I expect them to be playing with the agent locally, just like the customer that requested this, and 2) if options break the execution of Semgrep, we print the full failed command with all its arguments, so debugging is easy.

### Why I'd be worried about an allowlist

The obvious one is that keeping the allowlist up-to-date is effort. It'd inevitably be unpredictable and frustrating for users to find they cannot pass a specific flag through.

But there's a lot more than that. As it stands, we only see a list of tokens and don't understand flag-value relationships; we don't know what's a flag and what's a value. A simple regex to check if the token starts with `-` would still have false positives thinking a value is a flag. To avoid spurious failures, we'd have to commit to using the same arg parser as semgrep, with the same configuration.

And even worse, I don't think even all that would protect against most failures. I think most failure states would be introduced not by the wrong option being used, but the wrong value being passed. `--jobs max` is not a valid setting. `--config -` looks like it'd allow piping rules from stdin but this isn't supported. `--lang`/`--include` would have to be allowlisted for various valid use cases, but passing the wrong setting would break everything.

To sum up: to add more and more guardrails is a lot of effort and a lot more maintenance responsibility, and yet we cannot hope to get even just close to 100% safety. I think the best we could do is preventing around a third of custom option-induced errors.

### What next

If you agree with my reasoning, I still have some ideas:

1a. We could hide this option from `semgrep-agent --help`
1b. Or add a "warranty void" note to the flag's help description
2. We can add more help text in case of a `semgrep` run failure that points out that `--semgrep-opts` is likely why we failed when it's used, and asking that the user try removing these opts.

nbrahms:
@underyx Nice, this is a great write-up and makes a lot of sense to me. I like the combination of 1a and 2 personally, but happy to support any option you choose 👌

nbrahms (dismissed):

underyx:
```
=== looking for current issues in 3 files

=== failed command's STDOUT:

usage: semgrep [-h] [-g | -f CONFIG | -e PATTERN] [-l LANG] [--validate]
               [--strict] [--exclude EXCLUDE] [--include INCLUDE]
               [--no-git-ignore] [--skip-unknown-extensions]
               [--dangerously-allow-arbitrary-code-execution-from-rules]
<trimmed>

=== [ERROR] `/Users/underyx/.local/bin/semgrep --skip-unknown-extensions --disable-nosem --json --config https://semgrep.dev/c/p/r2c --help src/semgrep_agent/main.py src/semgrep_agent/semgrep.py stubs/sh/__init__.pyi` failed with exit code 0

You've passed these custom semgrep options: '--help'. Custom options can easily break the internals of semgrep-agent, so please reproduce this error without them before opening an issue.
```

This is the output now when failing with a custom option.

nbrahms (approved):
lgtm

PR #150, opened 2021-01-07
Fail with helpful error on no-rules

chmccreery:
A customer's CI was failing and they didn't know why

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -128,6 +128,20 @@ def main(
         click.echo(f"| using semgrep rules from {config}", err=True)
     elif sapp.is_configured:
         local_config_path, num_rules = sapp.download_rules()
+        if num_rules == 0:
+            message = """
+            == [ERROR] This policy will not run any rules
+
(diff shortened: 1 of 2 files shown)


nbrahms (approved):

src/semgrep_agent/main.py
```diff
@@ -128,6 +128,20 @@ def main(
...
+            Semgrep will only run rules in your policy that
+            have an action associated with them (notify or block).
+            We have a logging-only option coming soon, but in
+            the mean time, you can accomplish this by selecting
+            "notify" on the policy tab and not configuring any
+            channels on the notifications tab
+            """
+            message = dedent(message).strip()
+            click.echo(message, err=True)
+            sys.exit(1)
```
nbrahms, line 144: Should this be `sys.exit(0)` instead?

chmccreery: I don't think so- if a user is running 0 rules, they are not actually running Semgrep, which introduces security risks if you think you are running rules. There was a discussion about this previously and we decided that it was a serious enough misconfiguration to fail.

PR #153, opened 2021-01-10
Revert "Allow passing any arg with --semgrep-opts (#147)"

### LINKED ISSUE #152: latest semgrep-agent never completes its run
Sometime between Thursday and Saturday, semgrep-agent started exceeding the 20-minute timeout in my CI system (Buildkite). Prior to this, it used to take anywhere from 30 seconds to 2 minutes.

I am using the `returntocorp/semgrep-agent:v1` docker image and the `v1` tag was updated with a new docker image yesterday just prior to my first failing build:
https://hub.docker.com/layers/returntocorp/semgrep-agent/v1/images/sha256-93d7382e52[…]0d0aaf8c42f48ff9decc28950974cf038d7a9a201d405?context=explore

Here is what I know:

1. All runs were 2 minutes or less until this past Thursday; now all runs are timing out.
2. All runs used semgrep 0.35.0 on Python 3.7.9. (That is, semgrep-agent on Thursday reported this version and the newer semgrep-agent today reports the same version, leading me to believe the issue is in semgrep-agent code rather than semgrep code.)
3. All runs used differential mode (--baseline-ref).
4. All runs were scanning a very small number of files, anywhere from zero to a couple dozen files being scanned. (The zero is due to how I have done some configuration; it is an item on my TODO list but not relevant to this issue.)
5. All runs used your standard docker image (returntocorp/semgrep-agent:v1) though of course the reference of that tag changes frequently.
6. I had only recently expanded the number of rulesets in my policy. I just tried reducing the policy size by going back to just the single r2c-ci ruleset (reducing rule count from 508 to 119) but no difference--semgrep-agent still timing out.

This same behavior occurs both in Buildkite and on the command-line when I run it locally. Also, my repository is open-source, so you can use my actual data to observe the problem.

The repo is here: https://github.com/chef/automate

And here are the bits from my Makefile:
```
SEMGREP_CONTAINER := returntocorp/semgrep-action:v1
SEMGREP_COMMON_PARAMS := -m semgrep_agent --publish-token ${SEMGREP_TOKEN} --publish-deployment ${SEMGREP_ID}
SEMGREP_REPO := --env SEMGREP_REPO_NAME=chef/automate
DOCKER_PARAMS := --volume $(realpath .):/automate --workdir /automate

semgrep: ## runs differential semgrep, checking only changes in the current PR, just as is done in CI
	docker run -it --rm --init $(DOCKER_PARAMS) $(SEMGREP_REPO) $(SEMGREP_CONTAINER) python $(SEMGREP_COMMON_PARAMS) --baseline-ref master
```

Implementation details:
1. When run on a git branch, there will be a non-zero number of files. When run on master, though, there will be zero files. But either way, the same problem occurs--semgrep-agent gets stuck. 
2. In Buildkite, semgrep-agent times out at about 20 minutes; on the command-line, I have been letting it run while I wrote this up, so it will not time-out. Thus far, it has run for 40 minutes and is still running. This is on a branch with 32 (rather ordinary) files.

Priority: High
I have had to completely disable semgrep-agent until this can be resolved.

chmccreery:
This reverts commit 550275f02a71b675773d108f2f08c98658ce87cf.

(customer reported scans suddenly timing out after this commit)

I already pushed v1 tag back to before this commit and confirmed that this fixed the issue: https://r2c-community.slack.com/archives/G01CKAQDRA4/p1610212521012200

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -57,11 +57,10 @@ def get_aligned_command(title: str, subtext: str) -> str:
 )
 @click.option("--publish-token", envvar="INPUT_PUBLISHTOKEN", type=str)
 @click.option("--publish-deployment", envvar="INPUT_PUBLISHDEPLOYMENT", type=int)
+@click.option("--json", "json_output", hidden=True, is_flag=True)
 @click.option(
     "--gitlab-json", "gitlab_output", envvar="SEMGREP_GITLAB_JSON", is_flag=True
 )
-@click.option("--json", "json_output", hidden=True, is_flag=True)
-@click.option("--semgrep-opts", hidden=True, type=str)
 def main(
     config: str,
     baseline_ref: str,
@@ -70,7 +69,6 @@ def main(
     publish_deployment: int,
     json_output: bool,
     gitlab_output: bool,
-    semgrep_opts: str,
 ) -> NoReturn:
     click.echo("=== detecting environment", err=True)
     click.echo(
@@ -177,7 +175,6 @@ def main(
         meta.head_ref,
         semgrep.get_semgrepignore(sapp.scan.ignore_patterns),
         sapp.is_configured,
-        semgrep_opts=semgrep_opts,
     )
     new_findings = results.findings.new
 
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -1,14 +1,12 @@
 import io
 import json
 import os
-import shlex
 import sys
 import time
 import urllib.parse
 from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import datetime
-from json.decoder import JSONDecodeError
 from pathlib import Path
 from textwrap import dedent
 from textwrap import indent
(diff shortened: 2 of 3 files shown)


underyx:
I'm inclined to believe that this wasn't the reason for the timeouts. Note that around a whole day passed during the debugging process so there couldn't been something else causing and then later not-causing it 🤔 

Did a quick second read-through of the changes but still can't find why this might happen. I can test it a bit more locally later.

chmccreery:
Yeah, I didn't understand why this would have caused timeouts either, but it seemed like the easiest thing to test and then seemed to work. If we get _really_ stuck, we can try merging it back in and see if the issue returns, in which case we would know pretty conclusively that this was the issue but I just asked on the thread if anything else had changed between when it was broken and was fixed, and he said no. See linked issue for more information about the broken environment.

underyx (approved):

msorens:
Just wanted to clarify one point: the _residual_ effect was a timeout in my CI system, but the _actual_ characterization seems more like an infinite loop: when run manually (without the timeout constraint of my CI system), the command never finished (I aborted after 40 minutes or so).

PR #166, opened 2021-01-19
Add integration testing

### LINKED ISSUE #163: Add Test Suite for Regression Testing
After a high-priority issue with msorens of Chef was filed (https://github.com/returntocorp/semgrep-action/issues/152), follow-up discussions led to msorens stating that he disabled semgrep CI until further updates about this issue. 

We were unable to reproduce the issue, but reverted the commit that caused it, which seems to have fixed the problem. However, msorens is unsure about semgrep-actions and will disable semgrep CI until this issue is fixed. 

We should add in a test suite for semgrep-action in order to add repositories and tests such as these to let users know issues have been fixed and reassure them that these issues will not pop up again.

nbrahms:
Shameless reuse of bento infrastructure

Closes #163

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,7 @@ jobs:
       - id: setup-package-managers
         name: Setup package managers
         run: |
-          python -m pip install --upgrade poetry pipx
+          python -m pip install --upgrade poetry==1.1.4 pipx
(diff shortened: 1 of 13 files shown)


underyx (approved):
Okay with me as soon as tests pass.

dlukeomalley:
Long live Bento! 🙌

PR #168, opened 2021-01-19
Stop trying to find branch-off point on push events

underyx:


src/semgrep_agent/meta.py
```diff
@@ -238,7 +238,8 @@ def pr_title(self) -> Optional[str]:
         return str(pr_title) if pr_title else None
 
     def initialize_repo(self) -> None:
-        self._find_branchoff_point()
+        if self.event_name == "pull_request" and self.head_ref is not None:
```
underyx, line 241: Same condition as in normal usage: https://github.com/returntocorp/semgrep-action/pull/168/files#diff-5e09ede35c63d2b3762891167cd7d23ac07a4a93d835739c589675e1df77bbc5R205

nbrahms (approved):

PR #169, opened 2021-01-19
Allow passing any arg with --semgrep-opts (#147)

underyx:
Tested by running:
python -m semgrep_agent --config p/r2c --baseline-ref HEAD~3 \
--semgrep-opts '--max-lines-per-finding 45 -j "2" --timeout 90'


underyx:
Current test failure is because command output contains the failed semgrep command when semgrep errors out. This command has the target paths listed in random or platform-dependent order. @nbrahms ideas how to fix this?

PR #170, opened 2021-01-20
Use semgrep -o to ensure always parsable json

### LINKED ISSUE #159: Agent is repeatedly logging incrementing numbers of 'issues found'
```
| No current issues found
| 4 ignored issues found
| 15 current issues found
| 4 ignored issues found
| 76 current issues found
| 14 ignored issues found
| 79 current issues found
| 14 ignored issues found
| 79 current issues found
| 14 ignored issues found
| 79 current issues found
| 16 ignored issues found
| 89 current issues found
| 18 ignored issues found
| 94 current issues found
| 20 ignored issues found
| 111 current issues found
| 20 ignored issues found
| 122 current issues found
| 22 ignored issues found
| 127 current issues found
| 24 ignored issues found
```

Shared by @msorens on Slack: https://r2c-community.slack.com/archives/CK86BJ5DW/p1610467285371800

### LINKED ISSUE #162: semgrep-app should never fail to parse output of semgrep as json
Parsing of semgrep output in semgrep-action is fragile and sometimes breaks if any stray print statement from semgrep-core is output.

Quick v0 solution is to use the -o option for semgrep to output to a file and parse said file. This file is guaranteed to be parsable json without random print statments.

A better solution will be to have a way to call semgrep as a library/API to return json.

brendongo:
Semgrep output can sometimes be polluted causing json parsing to fail
Use -o option and parse the file so json parsing will never fail.

### DIFF
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -229,12 +229,19 @@ def invoke_semgrep(semgrep_args: List[str], targets: List[str]) -> Dict[str, Lis
     for chunk in chunked_iter(targets, PATHS_CHUNK_SIZE):
         with tempfile.NamedTemporaryFile("w") as output_json_file:
             args = semgrep_args.copy()
-            args.extend(["-o", output_json_file.name])
+            args.extend(
+                [
+                    "-o",
+                    output_json_file.name,  # nosem: python.lang.correctness.tempfile.flush.tempfile-without-flush
+                ]
+            )
             for c in chunk:
                 args.append(c)
 
             _ = semgrep_exec(*(args))
-            with open(output_json_file.name) as f:
+            with open(
+                output_json_file.name  # nosem: python.lang.correctness.tempfile.flush.tempfile-without-flush
+            ) as f:
                 parsed_output = json.load(f)
 
             output["results"].extend(parsed_output["results"])
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -2,6 +2,7 @@ import io
 import json
 import os
 import sys
+import tempfile
 import time
 import urllib.parse
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


tests/acceptance/semgrep-rules-repo/local-config-full-scan.err
```diff
@@ -10,9 +10,5 @@ Note that no baseline-ref was passed and detected environment is not Github or G
 === looking for current issues in 1473 files
 | 2 current issues found
 | No ignored issues found
-| 2 current issues found
```
brendongo, line 13: Old code was printing per chunk instead of once in the end

dlukeomalley: @brendongo Does this address https://github.com/returntocorp/semgrep-action/issues/159?

brendongo: Oh yes it does. Didn't realize there was an open issue

underyx: I was hoping we could keep that behavior in some form and just add lines saying "Processing batch 4 of 8" in-between, so that the agent doesn't spend minutes at a time with no output. But this seems difficult with this new architecture so might as well do it this way 👍

dlukeomalley: Yay, Christmas CAN come early 🙌

src/semgrep_agent/semgrep.py
```diff
@@ -211,13 +211,45 @@ def invoke_semgrep(
...
+        with tempfile.NamedTemporaryFile("w") as output_json_file:
+            args = semgrep_args.copy()
+            args.extend(
+                [
+                    "-o",
+                    output_json_file.name,  # nosem: python.lang.correctness.tempfile.flush.tempfile-without-flush
+                ]
+            )
+            for c in chunk:
+                args.append(c)
```
underyx, line 239:
```suggestion
            args = [*semgrep_args, "-o", output_json_file.name, *chunk]
```

underyx (approved):

PR #173, opened 2021-01-25
Add audit mode

underyx:
This is needed for the corresponding auto-install setting to work correctly.


README.md
```diff
@@ -109,6 +109,14 @@ commit your own `.semgrepignore`.
 Note that `.semgrepignore` is picked up only by the action,
 and will not be honored when running `semgrep` manually.
 
+### Silent Mode
```
dlukeomalley: @underyx @pabloest During testing today "silent" proved to be a confusing term our user and it wasn't clear from the title if meant it also succeeded if there were errors. Two ideas:

1. This feature could be renamed, perhaps to "audit" or some other indicator of being run in the background without disturbing the dev flow
2. Not configuring this in the workflow file. As is it reads as a duplicate capability to running a policy that doesn't block. I worry about user's trying to debug the lack of a breaking main/develop build when their policy says to block, but this overrides that behavior (and was set long ago when they first added Semgrep and have since forgotten). Could this capability be downscoped for now and considered as part of @chmccreery's work or a corollary to @brendongo's fail open feature?

underyx:
@dlukeomalley does the current "audit" content look good to you?

README.md
```diff
@@ -109,6 +109,14 @@ commit your own `.semgrepignore`.
 Note that `.semgrepignore` is picked up only by the action,
 and will not be honored when running `semgrep` manually.
 
+### Audit Mode
+
+You can set `--audit-on event_name` to always report success,
+even if some findings were found.
+For example, in GitHub Actions, if you set the option `auditOn: push`,
+scans will run but not fail on push events.
+You can set multiple event names with this option.
```
dlukeomalley: What event types are possible or supported? If these are GitHub specific, may we link to their docs?

README.md
```diff
@@ -109,6 +109,14 @@ commit your own `.semgrepignore`.
 Note that `.semgrepignore` is picked up only by the action,
 and will not be honored when running `semgrep` manually.
 
+### Audit Mode
```
dlukeomalley: Looks much better! Why might someone want to use this feature and/or what outcome does it allow them to achieve? I'd love to articulate that answer here for readers.

nbrahms (approved):

PR #183, opened 2021-02-02
Revamp Docker release job

underyx:


underyx:
Somehow this this new buildx thing also shaved one minute off of build times it seems.

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+          CUSTOM_TAGS=""
+          if [[ $GITHUB_REF == refs/tags/v1 ]]; then
+            CUSTOM_TAGS="stable"
+          fi
+          echo ::set-output name=custom_tags::${CUSTOM_TAGS}
+      - name: Docker meta
+        id: docker_meta
+        uses: crazy-max/ghaction-docker-meta@v1.11.0
+        with:
+          images: returntocorp/semgrep-action,returntocorp/semgrep-agent
```
brendongo, line 33: Are we pushing to two images?

brendongo (approved):

.github/workflows/release.yml
underyx: We had semgrep-action since the start and started publishing to -agent when expanding beyond GitHub Actions.

PR #191, opened 2021-02-04
Fix git log of scanned commits

### LINKED ISSUE #190: semgrep-action git log is misleading
Consider the following git graph:

```
* 5dded71 (branch-b) fix a bad thing
* 5da3731 a change
* 9624a9b two findings
| * 6ebfa1d (branch) another bad file
|/
* 20d0d15 bar
```

When running semgrep-agent on `branch-b` with baseline ref `branch`, we expect to see four commits displayed as the "commits being scanned". However, we instead see:

```
| scanning only the following commits:
|   * 5dded71 fix a bad thing
|   * 5da3731 a change
|   * 9624a9b two findings
| looking at 2 changed paths
```

nbrahms:
Previously we failed to log any commits between the
merge base and the baseline ref. Fix that.

For the following git graph:

```
* 5dded71 (branch-b) fix a bad thing
* 5da3731 a change
* 9624a9b two findings
| * 6ebfa1d (branch) another bad file
|/
* 20d0d15 bar
```

Running semgrep-agent on branch-b with

```
semgrep-agent --baseline-ref branch --config p/r2c-ci
```

would print:
```
| scanning only the following commits:
|   * 5dded71 fix a bad thing
|   * 5da3731 a change
|   * 9624a9b two findings
```

After this change, we now print:
```
| reporting findings introduced by these commits:
|   * 5dded71 fix a bad thing
|   * 5da3731 a change
|   * 9624a9b two findings
| also reporting findings fixed by these commits from the baseline branch:
|   * 6ebfa1d another bad file
| to exclude these latter commits, run with
|   --baseline-ref $(git merge-base branch head)
```

Fixes #190.


underyx (changes requested):
I think this is still misleading. After this PR we say "scanning […] `6ebfa1d another bad file`" which would be taken to mean that semgrep-agent would report any findings that first showed up in `6ebfa1d another bad file`. But that's not what happens, semgrep-agent reports findings that got removed in `6ebfa1d another bad file`. So it's essentially as if we were scanning the inverse of `6ebfa1d another bad file` in this case. Not sure how to represent that in the logs, it sounds kinda crazy that this even happens 😕

underyx:
So accurate log output would say

```
| reporting findings ADDED by the following commits:
|   * 5da3731 a change
|   * 9624a9b two findings
|   * 6ebfa1d another bad file
| and reporting issues FIXED by the following commits:
|   * 5dded71 fix a bad thing
```

underyx:
```
| reporting findings ADDED by the following commits:
|   * 5da3731 a change
|   * 9624a9b two findings
|   * 6ebfa1d another bad file
| and reporting issues FIXED by the following commits:
|   * 5dded71 fix a bad thing
| if you don't want to be notified about the latter group, please use
|   --baseline-ref $(git merge-base master HEAD)
```

underyx (approved):

PR #208, opened 2021-02-24
Improve error message when deployment ID is set but token is unset

underyx:


src/semgrep_agent/main.py
```diff
@@ -194,13 +194,33 @@ def main(
             "| using semgrep rules from the committed .semgrep/ directory", err=True
         )
         config = ".semgrep/"
+    elif publish_deployment is not None and not publish_token:
+        token_state = (
+            # can't check without hardcoding the environment variable name
+            # https://github.com/pallets/click/issues/1790
+            "set to '' (an empty string)"
+            if os.getenv("INPUT_PUBLISHTOKEN") == ""
```
nbrahms, line 202: I think this is correcterer:

```suggestion
            if publish_token == ""
```

nbrahms: Although honestly I would just fail on empty strings where we check the token length 🤷

src/semgrep_agent/main.py
```diff
@@ -194,13 +194,33 @@ def main(
...
+            You tried logging in as deployment ID #{publish_deployment}, but the deployment's API token is {token_state}.
+            If you're using a CI secret management feature to set it,
+            please ensure that your token secret (commonly named SEMGREP_APP_TOKEN) is available to this CI job.
+
+            You can find more details about authentication at
+            https://semgrep.dev/docs/semgrep-ci/#connecting-to-semgrep-app
+        """
+        message = dedent(message).strip()
+        click.echo(message, err=True)
+        sys.exit(1)
```
nbrahms, line 216: Question:

Do we adopt the same exit-code methodology in semgrep-app as we do in semgrep CLI? If so, this should be a `2`.

src/semgrep_agent/main.py
```diff
@@ -194,13 +194,33 @@ def main(
...
     else:
         message = """
             == [ERROR] you didn't configure what rules semgrep should scan for.
 
-            Please either set a config in the CI configuration according to
-            https://github.com/returntocorp/semgrep-action#configuration
-            or commit your own rules at the default path of `.semgrep.yml`
+            Please set a config in the CI configuration according to
```
nbrahms, line 221: Suggestion:

```suggestion
            Please set a config in {publish_url}/manage/policies according to
```

I think a direct link is more likely to lead to customer success.

src/semgrep_agent/main.py
```diff
@@ -194,13 +194,33 @@ def main(
             "| using semgrep rules from the committed .semgrep/ directory", err=True
         )
         config = ".semgrep/"
+    elif publish_deployment is not None and not publish_token:
```
nbrahms, line 197: Suggestion:

```suggestion
    elif publish_deployment and not publish_token:
```

Not sure exactly what we want to do when someone has an empty deployment ID, but this might be more robust?

src/semgrep_agent/main.py
```diff
@@ -194,13 +194,33 @@ def main(
...
+            "set to '' (an empty string)"
+            if os.getenv("INPUT_PUBLISHTOKEN") == ""
+            else "unset"
+        )
+        message = f"""
+            == [ERROR] you didn't set a token for authentication to semgrep.dev.
+
+            You tried logging in as deployment ID #{publish_deployment}, but the deployment's API token is {token_state}.
+            If you're using a CI secret management feature to set it,
+            please ensure that your token secret (commonly named SEMGREP_APP_TOKEN) is available to this CI job.
```
nbrahms: Nit:

Wrapping is weird here. One suggestion is to simply print a single line and let the terminal / display wrap for you.

nbrahms (approved):
Nice improvement @underyx . I left some minor thoughts below, but we could also merge as-is if you prefer.

src/semgrep_agent/main.py
underyx: This is what I tried first, but it doesn't work: https://github.com/pallets/click/issues/1790

src/semgrep_agent/main.py
underyx: I'm not sure, I followed the existing convention from the failure state right below this one.

src/semgrep_agent/main.py
underyx: This message is referring to offline mode, where people specify a `--config` setting in their CI configuration file.

src/semgrep_agent/main.py
underyx: If someone tries to set it as empty, click will convert it to `None`. With your suggestion deployment ID #0 (which shouldn't happen, granted) would confusingly be interpreted as unset and we'd fall through to assuming they were not trying to log in at all.

src/semgrep_agent/main.py
underyx:
```suggestion
            You tried logging in as deployment ID #{publish_deployment}, but the deployment's API token is {token_state}.
            If you're using a CI secret management feature to set it, please ensure that your token secret (commonly named SEMGREP_APP_TOKEN) is available to this CI job.
```

src/semgrep_agent/main.py
nbrahms: Right on, makes sense

src/semgrep_agent/main.py
nbrahms: How do you know if the user ran `--publish_token ${MY_SECRET}` if `MY_SECRET=""` ?

src/semgrep_agent/main.py
nbrahms: Hmm may be some tech debt here cc. @brendongo

src/semgrep_agent/main.py
underyx: They'd get an error from click saying `--publish-token` requires a value to be set.

src/semgrep_agent/main.py
nbrahms: ah.

src/semgrep_agent/main.py
nbrahms: Perfect.

underyx: Added a test case for this

tests/acceptance/misconfigured-generic/commands.yaml
```diff
@@ -15,14 +15,25 @@ steps:
...
+    expected_out: empty-deployment-token-env.out
+    expected_err: empty-deployment-token-env.err
+  - name: empty token via flag
+    description: >
+      run agent with deployment ID but token set to empty string in the command,
+      such as when a CI secret doesn't propagate correctly
+    command:
+      - semgrep-agent
+      - --publish-deployment=0
+      - --publish-token
```
nbrahms, line 36: I think you need:
```
- --publish-token
- ""
```

PR #211, opened 2021-02-24
Make diff-aware scans work with pull_request_target event

underyx:


underyx:
You can see a test run with `on: pull_request_target` here: https://github.com/returntocorp/semgrep-action/pull/210/checks?check_run_id=1972316671

brendongo (approved):

brendongo (commented):
Might be cleaner to have a function that is "is_pull_request_event" instead of returning a string fwiw

underyx:
> Might be cleaner to have a function that is "is_pull_request_event" instead of returning a string fwiw

Agree, rewrote it like that.

PR #218, opened 2021-03-04
Send fixed lines to Semgrep App if opted in

underyx:


src/semgrep_agent/main.py
```diff
@@ -257,9 +257,7 @@ def main(
 
     if json_output:
         # Output all new findings as json
-        json_contents = [
-            f.to_dict(omit=constants.PRIVACY_SENSITIVE_FIELDS) for f in new_findings
-        ]
+        json_contents = [f.to_dict(omit=set()) for f in new_findings]
```
chmccreery, line 260: Do you need the omit=set() here? I would guess it works the same without that

src/semgrep_agent/constants.py
```diff
@@ -3,6 +3,6 @@
 
 SUPPORT_EMAIL = "support@r2c.dev"
 
-PRIVACY_SENSITIVE_FIELDS = {"syntactic_context"}
+PRIVACY_SENSITIVE_FIELDS = {"syntactic_context", "fixed_lines"}
```
chmccreery, line 6: nice foresight adding this here!

src/semgrep_agent/main.py
chmccreery: or is it because by default to_dict was rewritten to use PRIVACY_SENSITIVE_FIELDS? Probably worth adding a comment explaining that

src/semgrep_agent/semgrep_app.py
```diff
@@ -153,14 +153,19 @@ def report_failure(self, error: SemgrepError) -> int:
     def report_results(self, results: Results) -> None:
         debug_echo(f"=== reporting results to semgrep app at {self.url}")
 
+        fields_to_omit = constants.PRIVACY_SENSITIVE_FIELDS.copy()
+
+        if "pr-comment-autofix" in os.getenv("SEMGREP_AGENT_OPT_IN_FEATURES", ""):
```
chmccreery, line 158: I am having trouble imaging a world where a user does not want to opt in here - should we just make it leave autofix recommendations any time PR comments are enabled?

underyx: This was because we promised not to collect customer source code, and I just remembered I still needed to clear this change with people.

src/semgrep_agent/semgrep_app.py
```diff
@@ -153,14 +153,19 @@ def report_failure(self, error: SemgrepError) -> int:
...
+            fields_to_omit.remove("fixed_lines")
+
         response: Optional["requests.Response"] = None
 
         response = self.session.post(
             f"{self.url}/api/agent/scan/{self.scan.id}/findings",
             json={
                 "token": os.getenv("GITHUB_TOKEN"),
```
chmccreery, line 167: Also perhaps add a comment here that this is now just a backup method of authentication, only used if github.get_install_token fails

chmccreery (approved):
Lgtm

src/semgrep_agent/semgrep_app.py
chmccreery: Oh, I thought we were ok as long as we didn't store their code anywhere - is this different than what we output to them in the terminal?

underyx: I worry that anything from this JSON blob is saved as findings data. We'll need to strip this in the backend to prevent it from being stored, and even after that we should clear this with someone with a better understanding of privacy policies.

src/semgrep_agent/main.py
underyx: yeah, that's what I did, but I'm not sure; if we had to add a comment to all its usages, then we're better off making it a required keyword argument with no default

do you think this confused you only because you were familiar with the old behavior, or would it be confusing to anyone reading this code for the first time?

src/semgrep_agent/semgrep_app.py
underyx:
```suggestion
                "token": os.getenv("GITHUB_TOKEN"),  # backup token in case the app is not available
```

src/semgrep_agent/main.py
chmccreery: I think the only reason it _didn't_ confuse me was because the new to_dict() definition was right above it in the git diff. What do you think of making the required argument be a boolean for "omit_sensitive" or something like that? I understand why you made the default behavior the safe thing here, and from a security perspective that seems right, but having a default value of type set that is not empty-set for a function where you had to override the definition is just counterintuitive to me... Not blocking, though.

underyx: Yeah, you're right. It would also be pretty easy to accidentally set your own one field to omit without keeping the sensitive fields' omission. But I also have too much time pressure right now so I'll just merge it as it is. :(

PR #219, opened 2021-03-09
Fix baseline checkout when git diff reports a directory

underyx:
Couldn't reproduce the original issue, so this is a blind and untested
fix for now.

Fixes https://github.com/returntocorp/semgrep-app/issues/1817


underyx:
Superseded by #226

PR #226, opened 2021-03-18
Ignore symlinks to directories

### LINKED ISSUE #227: semgrep action crashes when the git diff includes a symlink to a directory
# Manifestation
During clean up to a scan, semgrep-action looked like it was trying to call unlink on a directory. Semgrep-action does unlinking to remove Git Added files when trying to get to a baseline context.

It looked like somehow we were trying to remove a directory that wasn't even in the git diff (1 directory was not touched 2 git diff only shows files not directories). Turns out we got this behavior when there is a symlink to a directory that was committed.

# Reproduction
```
git init
mkdir foo && echo "2 == 2" > foo/bar.py && git add foo/bar.py && git commit -m 'add finding'
mkdir bar && touch bar/baz.py && git add bar/baz.py && git commit -m 'add no finding'
cd bar && ln -s '../foo' 'foo' && git add foo && git commit -m "Add symlink to dir" && ls -a
semgrep-agent --baseline-ref HEAD~1 --config p/r2c-ci
```

# Desired behavior:
Semgrep should not crash when there is a symlink to a directory.

brendongo:
Closes https://github.com/returntocorp/semgrep-action/issues/227


src/semgrep_agent/targets.py
```diff
@@ -124,22 +126,32 @@ def get_git_status(self) -> GitStatus:
...
+                    removed.append(resolved_name)
+                    fname = status_output[2]
+                    trim_size += 1
+                    added.append(resolved_name)
+                if code == StatusCode.Added:
+                    added.append(resolved_name)
+                if code == StatusCode.Modified:
+                    modified.append(resolved_name)
+                if code == StatusCode.Deleted:
+                    removed.append(resolved_name)
```
underyx, line 154:
```suggestion
                continue

            # The following detection for unmerged codes comes from `man git-status`
            if code == StatusCode.Unmerged:
                unmerged.append(resolved_name)
            if (
                code[0] == StatusCode.Renamed
            ):  # code is RXXX, where XXX is percent similarity
                removed.append(resolved_name)
                fname = status_output[2]
                trim_size += 1
                added.append(resolved_name)
            if code == StatusCode.Added:
                added.append(resolved_name)
            if code == StatusCode.Modified:
                modified.append(resolved_name)
            if code == StatusCode.Deleted:
                removed.append(resolved_name)
```

just to get rid of a level of indentation

tests/acceptance/symlink-dir/commands.yaml
```diff
@@ -0,0 +1,27 @@
...
+    command:
+      - bash
+      - -c
+      - mkdir bar && touch bar/baz.py && git add bar/baz.py && git commit -m 'add no finding'
+  - name: create symlink
+    description: Create directory with no finding
+    command:
+      - bash
+      - -c
+      - cd bar && ln -s '../foo' 'foo' && git add foo && git commit -m "Add symlink to dir" && ls -a
```
underyx, line 20: 😍

underyx (approved):
oh, forgot to leave the approval 😅

src/semgrep_agent/targets.py
brendongo: Continue will actually cause an infinite loop cause the loop condition needs to be updated

src/semgrep_agent/targets.py
```diff
@@ -124,22 +126,32 @@ def get_git_status(self) -> GitStatus:
...
+                # The following detection for unmerged codes comes from `man git-status`
+                if code == StatusCode.Unmerged:
+                    unmerged.append(resolved_name)
+                if (
+                    code[0] == StatusCode.Renamed
+                ):  # code is RXXX, where XXX is percent similarity
+                    removed.append(resolved_name)
+                    fname = status_output[2]
+                    trim_size += 1
+                    added.append(resolved_name)
```
karoun, line 148: 🐛 This is breaking PRs with renamed files:
```suggestion
                    added.append(self._fname_to_path(repo, fname))
```

karoun: https://github.com/returntocorp/semgrep-action/pull/233

chmccreery: 🙏

PR #231, opened 2021-03-19
Cm/small messaging improvements

chmccreery:


src/semgrep_agent/meta.py
```diff
@@ -339,7 +342,10 @@ def generate_meta_from_environment(baseline_ref: Optional[str]) -> GitMeta:
     else:  # nosem
         if not baseline_ref:
             click.echo(
-                "Note that no baseline-ref was passed and detected environment is not Github or Gitlab to will default to perfoming a full scan. If you want semgrep to only report on new findings from a pull request please set baseline-ref ",
+                get_aligned_command(
+                    "scan type",
+                    "full scan (if you want Semgrep to only report new findings, set baseline-ref)",
```
underyx:
```suggestion
                    "full scan (set --baseline-ref to report only new findings)",
```

underyx (approved):

src/semgrep_agent/meta.py
chmccreery: How about `full scan (set --baseline-ref to only scan changed files and report new findings)`?

underyx: @chmccreery updated my suggestion (still trying to keep it shorter)

PR #235, opened 2021-03-23
allow missing files to be removed

DrewDennison:


src/semgrep_agent/targets.py
```diff
@@ -303,6 +303,9 @@ def _baseline_context(self) -> Iterator[None]:
                 a.unlink()
             git.checkout(self._base_commit, "--", ".")
             yield
+        except FileNotFoundError:
+            click.echo(f"| {a} was not found when trying to delete", err=True)
+            yield
```
underyx: shouldn't this except clause go inside the for loop?

chmccreery (approved):

PR #241, opened 2021-03-29
Catch other exceptions besides ActionFailure and SemgrepError

chmccreery:
This looks weird (revert revert) because I accidentally pushed directly to develop but could not force-push to undo that accidental push.

So, I ended up reverting develop and reverting that revert in this pull request to get some eyes 😆 

I have tested that this does in fact catch a FileNotFoundError (original bug reported by user) and still gives enough information in the error message to be helpful to us when debugging (printing the error just gives None)

Screen Shot 2021-03-29 at 5 20 45 PM

@dlukeomalley after working on this today, I think that we should make fail_open an org-wide setting, so that we can catch _all_ semgrep-action errors and return 0, not just those that occur between when the scan is created and when it is finished (this PR).

Making fail-open org-wide requires some significant semgrep-app modifications, so I decided to just fix this first to downscope.

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import os
 import sys
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


chmccreery:
Fixes https://github.com/returntocorp/semgrep-app/issues/1922

nbrahms (approved):

PR #242, opened 2021-03-30
Add max Semgrep run time of 30 minutes in the action

nbrahms:
... to prevent 6 hour runs.

Also remove a bunch of unused imports.

30 minutes is at the top of our continuous distribution of scan times, and anything else really ought to be unacceptable.

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -248,6 +248,13 @@ def main(
     except SemgrepError as error:
         print_sh_error_info(error.stdout, error.stderr, error.command, error.exit_code)
         _handle_error(error.stderr, error.exit_code, sapp)
+    except sh.TimeoutException as error:
+        click.secho(
+            f"Semgrep took longer than {timeout} seconds to run; canceling this run",
+            err=True,
+            fg="red",
+        )
+        _handle_error(str(error), 2, sapp)
     except ActionFailure as error:
         click.secho(str(error), err=True, fg="red")
         _handle_error(str(error), 2, sapp)
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -107,9 +107,9 @@ def url(string: str) -> str:
 @click.option(
     "--timeout",
     envvar="SEMGREP_TIMEOUT",
-    default=300,
+    default=600,
     type=int,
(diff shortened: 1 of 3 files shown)


nbrahms:
Looks like this:

```
(semgrep-agent-kLEP9UGM-py3.9) ➜  generic git:(branch-b) semgrep-agent --config p/ci --timeout 5
| versions          - semgrep 0.41.1 on Python 3.9.2
| scan type         - full scan (set --baseline-ref to scan changed files only)
| environment       - running in environment git, triggering event is 'unknown'
| using semgrep rules from https://semgrep.dev/c/p/ci
| using default path ignore rules of common test and dependency directories
| found 8 files in the paths to be scanned
| skipping 2 files based on path ignore rules
=== looking for current issues in 6 files
Semgrep took longer than 5 seconds to run; canceling this run
```

This is inside the fail-open try/except block, so should obey the user's fail-open settings.

DrewDennison (approved):

underyx:
I'm happy with the 30 minute limit. I just wanted to note that waiting for only 5 minutes for the first scan's results caused autoinstall flow failures before, cause the first scan on the default branch can take longer than 5. But the 30 minute limit looks a lot nicer!

I also have a Redash query for getting the Nth percentile scan runtimes, in case you wanna set it to 99th percentile or something. Let me know if that'd help and I'll find it.

src/semgrep_agent/main.py
```diff
@@ -239,10 +243,18 @@ def main(
...
             semgrep.get_semgrepignore(sapp.scan.ignore_patterns),
             sapp.is_configured,
+            timeout=(timeout if timeout > 0 else None),
         )
     except SemgrepError as error:
         print_sh_error_info(error.stdout, error.stderr, error.command, error.exit_code)
         _handle_error(error.stderr, error.exit_code, sapp)
+    except sh.TimeoutException as error:
+        click.secho(
+            f"Semgrep took longer than {timeout} seconds to run; canceling this run",
```
dlukeomalley, line 253: What next steps can we recommend to the user? Where do I go to change this timeout, investigate why it was slow, or otherwise debug?

nbrahms:
Love the discussion here, but going to leave the outstanding items as follow-ups so we can get this out.

PR #258, opened 2021-04-15
Point users to relevant section of app for findings

adamberman:
Some semgrep users provide a deployment ID and auth token, allowing them
to send the results of a scan to the semgrep-app. For those users, let's
add text to the end of the output showing them where they can find those
results.

Output while scanning semgrep-action repo:

```
...
| No ignored issues found
=== not looking at pre-existing issues since there are no current issues
| to see your findings in the app, go to https://semgrep.dev/manage/findings?repo=semgrep-action
=== exiting with success status
```

This closes issue #256


nbrahms (approved):
lgtm

PR #266, opened 2021-04-22
Dockerfile: Install semgrep-agent command

underyx:


colleend (approved):
I, too, want to be like Bence when I grow up

colleend:
Should fix https://github.com/returntocorp/enterprise/issues/608

PR #268, opened 2021-04-28
Add .tox/ and *_test.go to default semgrepignore

underyx:


brendongo (approved):

src/semgrep_agent/templates/.semgrepignore
```diff
@@ -12,11 +12,13 @@ env/
...
 venv/
 .venv/
+.tox/
 *.min.js
 
-# Common test directories
+# Common test paths
 test/
 tests/
+*_test.go
```
brendongo, line 21: This seems risky in that we might ignore a file someone wants scanned but happened to name test?

underyx: my understanding is that the go ecosystem standardized on this naming so much, it's built into the `go` command itself; I imagine `go test` would complain if someone named a regular file ending in `_test.go` but I'm only 80% certain.

PR #278, opened 2021-05-13
Add retries on 413,429,500,501,503 errors

brendongo:
Old behavior was not retrying ton 50x errors. This adds those status_codes as something we want to retry with

### DIFF
--- a/src/semgrep_agent/semgrep_app.py
+++ b/src/semgrep_agent/semgrep_app.py
@@ -26,7 +26,12 @@ from semgrep_agent.utils import validate_publish_token
 
 # 4, 8, 16 seconds
 RETRYING_ADAPTER = requests.adapters.HTTPAdapter(
-    max_retries=Retry(total=3, backoff_factor=4, method_whitelist=["GET", "POST"])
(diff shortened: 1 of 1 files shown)


brendongo:
Doesn't fix https://github.com/returntocorp/semgrep-app/issues/2425 but at least mitigates affect on users

src/semgrep_agent/semgrep_app.py
```diff
@@ -26,7 +26,12 @@
 
 # 4, 8, 16 seconds
 RETRYING_ADAPTER = requests.adapters.HTTPAdapter(
-    max_retries=Retry(total=3, backoff_factor=4, method_whitelist=["GET", "POST"])
+    max_retries=Retry(
+        total=3,
+        backoff_factor=4,
+        method_whitelist=["GET", "POST"],
+        status_forcelist=(500, 502, 503),
```
underyx:
```suggestion
        status_forcelist=(413, 429, 500, 502, 503),
```

This readds two default retried statuses

underyx (approved):

PR #289, opened 2021-05-25
allow multiple configs

sabrinabrogren:
Fixes https://github.com/returntocorp/semgrep-app/issues/2597


README.md
```diff
@@ -66,15 +66,15 @@ This integration supports leaving inline PR comments via the Semgrep App. To lea
 
 ### Selecting Rules
 
-The `config` value lets you choose what rules and patterns semgrep should scan for.
-You can set specify rules in one of the following ways:
+The `config` values let you choose what rules and patterns semgrep should scan for.
```
sabrinabrogren: Is this enough of a change to make it clear that you can specify multiple configs?

chmccreery: Looks good to me. You could also do something like this if you want to be more explicit:
```suggestion
The `config` values (one or more) let you choose what rules and patterns semgrep should scan for.
```

src/semgrep_agent/main.py
```diff
@@ -67,7 +67,8 @@ def url(string: str) -> str:
     "--config",
     envvar="INPUT_CONFIG",
     type=str,
-    help="Define a rule, ruleset, or snippet used to scan this repository",
+    help="Define a rule, ruleset, or snippet used to scan this repository. You can pass in multiple `--config`s",
```
chmccreery, line 70: 👍

src/semgrep_agent/main.py
```diff
@@ -87,7 +88,7 @@ def url(string: str) -> str:
     "--publish-deployment",
     envvar="INPUT_PUBLISHDEPLOYMENT",
     type=int,
-    help="You semgrep.dev organization ID (requires --publish-token)",
+    help="Your semgrep.dev organization ID (requires --publish-token)",
```
chmccreery: Let's call this `deployment ID` instead of `organization ID` to match the new settings panel.
```suggestion
    help="Your semgrep.dev deployment ID (requires --publish-token)",
```

sabrinabrogren: Should this also change on line 85? Currently:
`Your semgrep.dev API token (only needed if specifying a publish organization)`

tests/acceptance/multiconfig-generic/base-log.err
```diff
@@ -0,0 +1 @@
+
```
chmccreery, line 1: Whats this?

sabrinabrogren: I copied all these files from the baseline generic test and just changed it so I could test with multiple configs! So I believe this is just the error file to check against (empty because there should be no errors).

tests/acceptance/multiconfig-generic/commands.yaml
```diff
@@ -0,0 +1,21 @@
...
+      - echo "2 == 2" > bar.py && git add bar.py && git commit -m 'add finding'
+  - name: new log
+    description: git log with commits
+    command: git log --stat
+    returncode: 0
+    expected_out: new-log.out
+    expected_err: new-log.err
+  - name: new agent
+    description: run agent on a single new file
+    command: semgrep-agent --baseline-ref HEAD~1 --config p/r2c-ci --config p/r2c
```
chmccreery, line 18: Yay adding tests! 🎉

src/semgrep_agent/main.py
chmccreery: Ooh, yeah, I think "publish deployment" or "deployment ID" instead of "publish organization" - having two terms is already confusing enough 😅

chmccreery (approved):

src/semgrep_agent/main.py
```diff
@@ -162,7 +162,7 @@ def main(
         click.secho(str(error), err=True, fg="red")
         _handle_error(str(error), 2, sapp, meta)
     except InvalidGitRepositoryError as error:
-        click.secho("Current directory is not a github repository", err=True, fg="red")
+        click.secho("Current directory is not a git repository", err=True, fg="red")
```
chmccreery, line 165: I had this fixup lingering on my local copy of the repo and didn't think it was worth its own PR so thought I would sneak it in here 🤷

chmccreery: If any tests fail or anything, we can just revert it - don't want to slow down this PR!

PR #291, opened 2021-05-26
Allow users to opt out of metrics with the --disable-metrics flag

chmccreery:
This PR adds the the ability for Semgrep CI to collect opt-out non-identifiable aggregate metrics for improving the user experience, guiding Semgrep feature development, and identifying regressions. It relies on Semgrep CLI’s metric collection, which is discussed in detail in that project’s [PRIVACY.md](https://github.com/returntocorp/semgrep/blob/develop/PRIVACY.md).

Users can opt-out of metrics by setting the environment variable `SEMGREP_SEND_METRICS=0` or Semgrep CI’s CLI flag `--disable-metrics`.

We have aggressive performance goals for Semgrep CI so capturing these metrics helps us understand real-world performance and be data-driven in prioritization.

The specific metrics collected by Semgrep CI, descriptions of fields, example values, and additional context are included in this project’s PRIVACY.md file (updated as part of this PR).

Changes:

- Run Semgrep CI with metrics enabled by default
- Allow users to opt out of metrics with the `--disable-metrics` flag (and `SEMGREP_SEND_METRICS` env variable)
- Never collect metrics on the baselining run, as this would double count

### DIFF
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -47,7 +47,7 @@ Scan data provide information on the environment and performance of Semgrep. The
 
 Findings data are used to provide human readable content for notifications and integrations, as well tracking results as new, fixed, or duplicate. The classes of data included are:
 
-- Check ID and metadata (as defined in the rule definition, e.g. OWASP category, message, severity)
+- Check ID and metadata (as defined in the rule definition; e.g. OWASP category, message, severity)
 - Code location, including file path, that triggered findings
 - A one-way hash of a unique code identifier that includes the triggering code content
 - Code content is not collected
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -12,11 +12,15 @@ Scan data provide information on the environment and performance of Semgrep. The
 - Scan metadata, including type of scan and scan parameters (e.g. files or branches to ignore)
 - Review and review-requester identifying data (e.g pull-request ID, branch, merge base, request author)
 - Semgrep environment (e.g. version, interpreter, timestamp)
+- Semgrep performance (scan duration and one-way hashes of rules that ran)
+- Project size in bytes
+- Errors (compile-time errors and return codes)
 
 ## Findings data
 
 Findings data are used to provide human readable content for notifications and integrations, as well tracking results as new, fixed, or duplicate. The classes of data included are:
 - Check ID and metadata (as defined in the rule definition; e.g. OWASP category, message, severity)
 - Code location, including file path, that triggered finding
+- A one-way hash of rule definitions that yield findings
 - A one-way hash of a unique code identifier that includes the triggering code content
 - **Code content is not collected**
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -22,7 +22,7 @@ Semgrep CI collects opt-out non-identifiable aggregate metrics for improving the
 
 ## Opt-out behavior
 
-Semgrep CI’s metrics can be disabled by setting the environment variable SEMGREP_SEND_METRICS=0 or using the flag --disable-metrics. If this environment variable or flag is not set, aggregate metrics are enabled.
+Semgrep CI’s metrics can be disabled by setting the environment variable `SEMGREP_SEND_METRICS=0` or using the flag `--disable-metrics`. If this environment variable or flag is not set, aggregate metrics are enabled.
 
 
 ## Data with Semgrep App only
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -12,9 +12,9 @@ Semgrep CI (also known as semgrep-action or semgrep-agent) may collect non-ident
 
 These principles inform our decisions around data collection:
 
-1. *Transparency*: Collect and use data in a way that is clearly explained to the user and benefits them
-2. *User control*: Put users in control of their data at all times
-3. *Limited data*: Collect what is needed, de-identify where possible, and delete when no longer necessary
+1. **Transparency**: Collect and use data in a way that is clearly explained to the user and benefits them
+2. **User control**: Put users in control of their data at all times
+3. **Limited data**: Collect what is needed, de-identify where possible, and delete when no longer necessary
 
 ## Collected data
 
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -27,7 +27,7 @@ Semgrep CI’s metrics can be disabled by setting the environment variable `SEMG
 
 ## Data with Semgrep App only
 
-For Semgrep App users running Semgrep CI with a SEMGREP_APP_TOKEN set, data is sent to power your dashboard, notification, and finding features. These data are ONLY sent when using Semgrep CI in an App-connected mode and are default-disabled for Semgrep CI users.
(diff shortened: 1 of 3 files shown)


PRIVACY.md
```diff
@@ -12,11 +12,15 @@ Scan data provide information on the environment and performance of Semgrep. The
...
+- Project size in bytes
+- Errors (compile-time errors and return codes)
 
 ## Findings data
 
 Findings data are used to provide human readable content for notifications and integrations, as well tracking results as new, fixed, or duplicate. The classes of data included are:
 - Check ID and metadata (as defined in the rule definition; e.g. OWASP category, message, severity)
 - Code location, including file path, that triggered finding
+- A one-way hash of rule definitions that yield findings
 - A one-way hash of a unique code identifier that includes the triggering code content
```
chmccreery: @pabloest please review

src/semgrep_agent/main.py
```diff
@@ -90,6 +90,12 @@ def url(string: str) -> str:
     type=int,
     help="Your semgrep.dev deployment ID (requires --publish-token)",
 )
+@click.option(
+    "--disable-metrics",
+    envvar="DISABLE_METRICS",
+    is_flag=True,
+    help="Disable anonymized metrics collection (such as scan duration), used to improve Semgrep",
```
chmccreery: @pabloest please review

PRIVACY.md
chmccreery: Note that we do not distinguish between logged-in users and not-logged-in users in the PRIVACY.md as of right now

src/semgrep_agent/main.py
```diff
@@ -90,6 +90,12 @@ def url(string: str) -> str:
     type=int,
     help="Your semgrep.dev deployment ID (requires --publish-token)",
 )
+@click.option(
+    "--disable-metrics",
```
nbrahms: The correct (TM) way to do this is

```suggestion
    "--enable-metrics/--disable-metrics",
    envvar="ENABLE_METRICS",
    default=False
```

see https://click.palletsprojects.com/en/8.0.x/options/#boolean-flags

nbrahms (changes requested):

nbrahms (approved):
lgtm

dlukeomalley:
@chmccreery @nbrahms Catching up, what led the choice in flags differing from those that Semgrep CLI uses [here](https://github.com/returntocorp/semgrep/blob/develop/PRIVACY.md#opt-in-behavior)? Would it be possible to just abide by Semgrep CLI environment variable naming?

I do see a case for using `--disable-metrics` as a flag instead of `--enable-metrics` given the opt-out nature of this feature.

What happens in CLI (or agent) when both are set, and they conflict?

chmccreery:
> @chmccreery @nbrahms Catching up, what led the choice in flags differing from those that Semgrep CLI uses [here](https://github.com/returntocorp/semgrep/blob/develop/PRIVACY.md#opt-in-behavior)? Would it be possible to just abide by Semgrep CLI environment variable naming?

I can easily change the name of the environment variable to be SEMGREP_SEND_METRICS instead of ENABLE_METRICS. I was following the convention of the other flags in semgrep-agent, where the env variable has the same name as the flag but all upper-cased with underscores. The env variable name doesn't show up in a call to `semgrep-agent --help` and this convention seemed like the expected behavior to me. 

> I do see a case for using `--disable-metrics` as a flag instead of `--enable-metrics` given the opt-out nature of this feature.

This was my thinking as well 🤷 But I also see the case for having both flags, as Nathan suggested, which makes it easier to switch the default behavior to opt-in in the future, if we wanted to.

> What happens in CLI (or agent) when both are set, and they conflict?

In both, the --flag takes precedence over the environment variable.

PRIVACY.md
```diff
@@ -1,22 +1,53 @@
...
+2. *User control*: Put users in control of their data at all times
+3. *Limited data*: Collect what is needed, de-identify where possible, and delete when no longer necessary
+
+## Collected data
+
+Semgrep CI collects opt-out non-identifiable aggregate metrics for improving the user experience, guiding Semgrep feature development, and identifying regressions. It relies on Semgrep CLI’s metric collection, which is discussed in detail in that project’s [PRIVACY.md](https://github.com/returntocorp/semgrep/blob/develop/PRIVACY.md).
+
+## Opt-out behavior
+
+Semgrep CI’s metrics can be disabled by setting the environment variable SEMGREP_SEND_METRICS=0 or using the flag --disable-metrics. If this environment variable or flag is not set, aggregate metrics are enabled.
```
dlukeomalley:
```suggestion
Semgrep CI’s metrics can be disabled by setting the environment variable `SEMGREP_SEND_METRICS=0` or using the flag `--disable-metrics`. If this environment variable or flag is not set, aggregate metrics are enabled.
```

dlukeomalley: @chmccreery We'll need to update this section with whatever environment variable and flag names you choose.

PRIVACY.md
```diff
@@ -1,22 +1,53 @@
...
+
+
+## Data with Semgrep App only
+
+For Semgrep App users running Semgrep CI with a SEMGREP_APP_TOKEN set, data is sent to power your dashboard, notification, and finding features. These data are ONLY sent when using Semgrep CI in an App-connected mode and are default-disabled for Semgrep CI users.
```
dlukeomalley:
```suggestion
For Semgrep App users running Semgrep CI with a `SEMGREP_APP_TOKEN` set, data is sent to power your dashboard, notification, and finding features. These data are ONLY sent when using Semgrep CI in an App-connected mode and are default-disabled for Semgrep CI users.
```

PRIVACY.md
```diff
@@ -1,22 +1,53 @@
...
+- the breakdown of the data that are and are not collected
+- how to opt-out of Semgrep CI’s metrics
+
+## Principles
+
+These principles inform our decisions around data collection:
+
+1. *Transparency*: Collect and use data in a way that is clearly explained to the user and benefits them
+2. *User control*: Put users in control of their data at all times
+3. *Limited data*: Collect what is needed, de-identify where possible, and delete when no longer necessary
```
dlukeomalley:
```suggestion
1. **Transparency**: Collect and use data in a way that is clearly explained to the user and benefits them
2. **User control**: Put users in control of their data at all times
3. **Limited data**: Collect what is needed, de-identify where possible, and delete when no longer necessary
```

Adding bold for consistency with styling on https://github.com/returntocorp/semgrep/blob/develop/PRIVACY.md

dlukeomalley (dismissed):

PRIVACY.md
chmccreery: I changed it to SEMGREP_SEND_METRICS to match the Semgrep CLI name 😄

chmccreery:
@pabloest where are we at with messaging users about this change?

pabloest:
@chmccreery We notified folks in the community Slack about this change last week on Thursday. One person asked a clarifying question there.

PR #296, opened 2021-06-02
Update CONTRIBUTING.md

chmccreery:
Update `make release` instructions to be clearer

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,7 +68,7 @@ them!
 
 ## Release
 
-Let CI pass on GitHub Actions before releasing. 
+Let CI pass on GitHub Actions before releasing.
 After you merge your PR, check out develop and pull to get your latest changes.
 The following command will change all action runs to use your current `HEAD`:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -69,11 +69,9 @@ them!
...
 Let CI pass on GitHub Actions before releasing.
+After you merge your PR, check out develop and pull to get your latest changes.
 The following command will change all action runs to use your current `HEAD`:
-```make release```
-
-### Publishing to pypi
 
 ```
-poetry publish
+make release
```
sabrinabrogren, line 76: Is `poetry publish` just no longer needed?

sabrinabrogren (approved):

CONTRIBUTING.md
chmccreery: I think so? That's why I tagged @brendongo for review, too. My memory is that this is not something we do anymore.

brendongo (approved):

PR #298, opened 2021-06-04
Improve messaging on what happens when config and policy are both present

chmccreery:
If users supply a config and also publishDeployment/publishToken to log in, this is likely an error, so let's message about it really loudly.

Screen Shot 2021-06-07 at 12 13 28 PM

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -235,14 +235,17 @@ def protected_main(
     click.echo("=== setting up agent configuration", err=True)
     if config:
         if sapp.is_configured:
-            click.secho(
-                f"\nDetected a config flag: no longer using rules configured on the web.",
-                fg="yellow",
-            )
-            click.echo(
-                f"If you wish to use Semgrep Cloud features, please remove the config flag\n"
-                "and add rules and rulesets to your policies instead.\n"
-            )
+            message = """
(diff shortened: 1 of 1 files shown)


dlukeomalley:
@chmccreery May we hard fail with a helpful error message in this case? For any "invalid" configuration combination we should be loud, louder than just printing to the job output. The case I'm worried about is that folks wont' read this output unless their job fails, so it's not loud in the current implementation.

In short, I'd like it to be impossible to unknowingly misconfigure Semgrep CI. In the future this behavior may change as we support local and remote configurations simultaneously.

chmccreery:
> @chmccreery May we hard fail with a helpful error message in this case? For any "invalid" configuration combination we should be loud, louder than just printing to the job output. The case I'm worried about is that folks wont' read this output unless their job fails, so it's not loud in the current implementation.
> 
> In short, I'd like it to be impossible to unknowingly misconfigure Semgrep CI. In the future this behavior may change as we support local and remote configurations simultaneously.

Yep, for sure! While everyone was out last week, I didn't want to change the behavior itself in case this was configured intentionally this way for a reason. I wasn't sure, for example, if this was intended behavior, or if the intended behavior was to use both the rules from the UI and the ones in the config if both were present, or something else. 

I will change to error message instead of warning here!

brendongo (approved):

PR #299, opened 2021-06-04
Allow users to override our default behavior of rewrite rule-ids

### LINKED ISSUE #290: The returntocorp/semgrep-agent does not include the “--no-rewrite-rule-ids” option
Currently we do not have the option to ignore findings while using the "semgrep-agent" in our CI pipeline.  Given the Rule IDs are being rewritten with a random CI directory name we are unable to add the Rule ID to a “nosemgrep: rule-id” inline comment.

Example renaming of a Rule ID:
```
tmp.tmp.ggDeOk.semgrep-rules.typescript.react.security.audit.react-missing-noopener.react-missing-noopener
```

Error message from semgrep-agent:
```
Error: no such option: --no-rewrite-rule-ids
```

chmccreery:
Allow users to override our default behavior of rewriting rule-ids when their rules are stored in a .semgrep/ directory, the same way they would be able to with semgrep CLI.

Fixes #290

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -329,7 +329,7 @@ def protected_main(
         meta.base_commit_ref,
         meta.head_ref,
         semgrep.get_semgrepignore(sapp.scan.ignore_patterns),
-        rewrite_rule_ids,
+        rewrite_rule_ids and not sapp.is_configured,
         enable_metrics,
         timeout=(timeout if timeout > 0 else None),
     )
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -102,7 +102,8 @@ def url(string: str) -> str:
     envvar="REWRITE_RULE_IDS",
     default=True,
     is_flag=True,
-    help="Specify whether semgrep should or should not rewrite rule-ids based on directory structure",
+    help="Specify whether semgrep should or should not rewrite rule-ids based on directory structure (MAY BE DEPRECATED ONCE ANY ARBITRARY SEMGREP CLI FLAG CAN BE PASSED IN)",
+    hidden=True,
 )
 @click.option(
     "--publish-url",
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -98,6 +98,13 @@ def url(string: str) -> str:
     help="Enable (default) or disable anonymized metrics used to improve Semgrep",
 )
 @click.option(
+    "--rewrite-rule-ids/--no-rewrite-rule-ids",
+    envvar="REWRITE_RULE_IDS",
+    default=True,
+    is_flag=True,
+    help="Specify whether semgrep should or should not rewrite rule-ids based on directory structure",
+)
+@click.option(
     "--publish-url",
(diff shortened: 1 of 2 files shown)


src/semgrep_agent/main.py
```diff
@@ -320,7 +329,7 @@ def protected_main(
         meta.base_commit_ref,
         meta.head_ref,
         semgrep.get_semgrepignore(sapp.scan.ignore_patterns),
-        sapp.is_configured,
+        rewrite_rule_ids and not sapp.is_configured,
```
brendongo, line 333: Should we allow this to be set to false if sapp.is_configured is true? i.e. should this be an or?

chmccreery: I don't think so - if they are using semgrep cloud and try to rewrite rule ids, semgrepignores will no longer work and rule names will be pre-pended with tmp.waf93 or something I think. Since all their rules will be in one file, there is strictly no benefit to having no-rewrite-rule-ids set.

chmccreery: Perhaps that means I should change the flag to not let them set --rewrite-rule-ids to true ever, but I didn't like how many double-negatives that put into the logic. It made it hard to reason about.

brendongo: So if we always want rules from semgrep-cloud to rewrite rule_ids shouldn't this be `rewrite_rule_ids or not sapp.is_configured`?

chmccreery: We do not always want rules from semgrep-cloud to rewrite rule_ids. We actually _never_ want rules from semgrep-cloud to rewrite rule_ids because rules are downloaded and saved to a flat config file - there is no meaningful path to prefix the rules with. The logic is:

if `sapp.is_configured` is true, then `not sapp.is_configured` is false, so anything anded with `not sapp.is_configured` is also false, so rule ids will never be rewritten ✅

chmccreery: Meanwhile, if `sapp.is_configured` is false, then `not sapp.is_configured` is true, so then we follow whatever is specified by the commandline flag ✅

brendongo: Ah makes sense. I got confused with all the nots and rewrite-not-rewrite :D

brendongo:
This looks good. i'm wondering if we should revisit https://github.com/returntocorp/semgrep-action/pull/169 since once we add this flag it will be involved to remove it later.

src/semgrep_agent/main.py
chmccreery: Yeah, it is confusing 🙁 but I couldn't figure out how to write this in a less-confusing way given how semgrep already does it

chmccreery:
> This looks good. i'm wondering if we should revisit #169 since once we add this flag it will be involved to remove it later.

Yeah, good to keep #169 in mind. I'll defer to @dlukeomalley on whether this bug fix is still high-priority or whether we should pause and not merge this until/unless #169 gets resolved 🤷

brendongo (approved):
Approving to unblock since it looks like https://github.com/returntocorp/semgrep-action/pull/169 needs more work still

PR #300, opened 2021-06-08
Update readme with new docs

underyx:


dlukeomalley (approved):
Approved conditionally on tests passing.

underyx:
@pabloest I'll go ahead and merge with Luke's approval but would still appreciate your review post-merge!

pabloest:
> @pabloest I'll go ahead and merge with Luke's approval but would still appreciate your review post-merge!

Sure thing!

I tried self-requesting a post-merge review but that didn’t work. I’ll just enter comments or questions in the GH comments below.

pabloest:
**Badges**

1. Do we need to periodically manually update the Slack badge number? I am happy to do so and would just set a reminder for myself to look into it every few weeks or so.
2. NICE on the docs badge

**Technical details > Packaging**
1. Thank you for keeping this section somewhere.
2. @dlukeomalley had suggested removing the reference to the Python package, but I’m not sure if you two had follow-on discussion about this.

**Technical details > Usage outside CI**
1. There’s excellent content here, but in the docs we had a bit of trouble finding the right home for it. I’m glad it’s here at least. How about breaking it out of “technical details” and making it a section of its own?

Overall I think this is awesome!

underyx:
> Do we need to periodically manually update the Slack badge number? I am happy to do so and would just set a reminder for myself to look into it every few weeks or so.

I think it can be a 12 month reminder if anything :D 700+ will stay technically true forever, and there is not a major difference between seeing 700+ and 1100+ I think 🤔

underyx:
> @dlukeomalley had suggested removing the reference to the Python package, but I’m not sure if you two had follow-on discussion about this.

He clicked on it with a grumble and then exerted a resigned but accepting sigh when realizing it links to a specific directory. I will consider that the follow-up discussion done.

pabloest:
Yes I was probably being optimistic about the Slack number :D

PR #308, opened 2021-06-16
Dockerfile: Bump cryptography version

underyx:
This PR removes the version pinning we had for poetry to avoid installing Rust, and just… installs rust instead. This way we can use the latest cryptography version.

The PR also bumps all python dependencies to their latest versions.

ruamel.yaml is replaced with PyYAML to reflect the same change we made in semgrep-app months ago. PyYAML has a larger user community and operates faster.

Download size of image went from 101.13 MB to 101.44 MB.


underyx:
@mschwager this addresses a comment you had on #288

brendongo (approved):

PR #314, opened 2021-06-28
Report searched paths

chmccreery:
This is one step toward having a more reliable 'fixed' status

By sending the list of searched paths to the semgrep-app backend, we enable semgrep-app to distinguish between issues that no longer show up because they were _fixed_ and issues that no longer show up because their paths are _ignored_. 

In a future PR we will actually use these data on semgrep-app to change such issues to status "REMOVED" rather than status "FIXED".

### DIFF
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -132,6 +132,9 @@ def get_findings(
             click.echo(
                 "=== looking for current issues in " + unit_len(paths, "file"), err=True
             )
+            if len(paths) < 100:
+                for path in paths:
+                    debug_echo(f"searching {str(path)}")
 
             args = [
                 "--skip-unknown-extensions",
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -126,7 +126,7 @@ def get_findings(
         rewrite_args = [] if rewrite_rule_ids else ["--no-rewrite-rule-ids"]
         metrics_args = ["--enable-metrics"] if enable_metrics else []
         debug_echo("=== seeing if there are any findings")
-        findings = FindingSets(ignored_paths=set(targets.ignored_paths))
+        findings = FindingSets(searched_paths=set(targets.searched_paths))
 
         with targets.current_paths() as paths:
             click.echo(
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -132,9 +132,9 @@ def get_findings(
             click.echo(
                 "=== looking for current issues in " + unit_len(paths, "file"), err=True
             )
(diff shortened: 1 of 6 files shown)


src/semgrep_agent/targets.py
```diff
@@ -70,11 +76,11 @@ class TargetFileManager:
     """
 
     _base_path = attr.ib(type=Path)
-    _paths = attr.ib(type=List[Path])
+    _all_paths = attr.ib(type=List[Path])
```
nbrahms, line 79: Suggestion:

Maybe `_input_paths` or something would make this clear that it's the user input to the tool?

src/semgrep_agent/semgrep.py
```diff
@@ -126,12 +126,15 @@ def get_findings(
...
         metrics_args = ["--enable-metrics"] if enable_metrics else []
         debug_echo("=== seeing if there are any findings")
-        findings = FindingSets()
+        findings = FindingSets(searched_paths=set(targets.searched_paths))
 
         with targets.current_paths() as paths:
             click.echo(
                 "=== looking for current issues in " + unit_len(paths, "file"), err=True
             )
+            if len(paths) < 100:
```
nbrahms: Suggestion: remove the length restriction here

Otherwise it's hard to tell what's going on; it should be fine to spew 10k lines of debug log

src/semgrep_agent/targets.py
```diff
@@ -210,21 +216,31 @@ def _get_target_files(self) -> List[Path]:
...
 
-        skipped_count = len(walked_entries) - len(filtered)
-        if skipped_count:
+        if ignored_paths:
             click.echo(
-                f"| skipping {unit_len(range(skipped_count), 'file')} based on path ignore rules",
+                f"| skipping {unit_len(ignored_paths, 'file')} based on path ignore rules",
                 err=True,
             )
+            if len(patterns) < 100:
```
nbrahms: Suggestion:

- Same comment on a length guard here
- Can we add more detail to what's being skipped and why?

E.g.:
```
debug_echo(f"Ignoring files matching pattern '{p}'")
```

src/semgrep_agent/targets.py
```diff
@@ -376,4 +392,13 @@ def current_paths(self) -> Iterator[List[Path]]:
         :return: A list of paths
         :raises ActionFailure: If git cannot detect a HEAD commit or unmerged files exist
         """
-        yield self._target_paths
+        yield self._paths.targeted
+
+    @property
+    def searched_paths(self) -> List[Path]:
+        """
+        Return list of paths that have been ignored.
```
nbrahms: Please change:

The function name, docstring, and actual returned data all disagree with each other

nbrahms (approved):
One change request

PR #327, opened 2021-07-07
Add debug logging and timeout to git checkout

nbrahms:
We sometimes see GitHub actions hanging on git checkout. Let's
- prevent a hang here by timing out (will throw an exception,
  causing the agent to crash) after 5 minutes
- add debug logging to show us what's going on

Also use sh's callback functionality to avoid spinning up an auxiliary
thread.

### DIFF
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -28,7 +28,6 @@ from semgrep_agent.findings import Finding
 from semgrep_agent.findings import FindingSets
 from semgrep_agent.targets import TargetFileManager
 from semgrep_agent.utils import debug_echo
-from semgrep_agent.utils import debug_file_descriptor
 from semgrep_agent.utils import get_git_repo
 from semgrep_agent.utils import is_debug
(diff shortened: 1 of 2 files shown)


src/semgrep_agent/semgrep.py
```diff
@@ -378,7 +376,8 @@ def _fix_head_for_github(
         if not stashed_rev:
             stashed_rev = git(["rev-parse", "HEAD"]).stdout.decode("utf-8").rstrip()
         click.echo(f"| not on head ref {head_ref}; checking that out now...", err=True)
-        git.checkout([head_ref])
+        git.checkout([head_ref], _timeout=300, _out=debug_echo, _err=debug_echo)
```
emjin, line 379: What are the units on this?

emjin (approved):

src/semgrep_agent/semgrep.py
nbrahms: seconds :)
https://amoffat.github.io/sh/sections/special_arguments.html#timeout

PR #328, opened 2021-07-09
specify --json flag via envvar

ronwaisberg-okta:
Hi there,

My use case for semgrep-action in GitHub Actions requires post-processing on the agent's output, ideally in JSON format. However the `--json` flag cannot be specified through the inputs defined in action.yml or an environment variable. This small PR allows specifying the flag through the `SEMGREP_JSON_OUTPUT` envvar.

Please let me know if there's anything I can change. Cheers!

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -113,7 +113,13 @@ def url(string: str) -> str:
     help="The URL of the Semgrep app",
     hidden=True,
 )
-@click.option("--json", "json_output", hidden=True, is_flag=True)
+@click.option(
+    "--json",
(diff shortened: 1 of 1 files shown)


underyx:
Makes sense and looks good to me! I'll merge & release this feature in 5-10 minutes.

underyx (approved):

ronwaisberg-okta:
@underyx I've linted my change to pass that test but some of the other workflows are failing likely because my PR is from a fork e.g. https://github.com/returntocorp/semgrep-action/pull/328/checks?check_run_id=3059298403

PR #338, opened 2021-07-20
Print and report semgrep errors

nbrahms:
Previously, we just ate our errors :(. Instead, let's surface them to
users AND upload them to our backend.

Also report correct exit code.

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -15,6 +15,9 @@ from git import InvalidGitRepositoryError
 
 from semgrep_agent import formatter
 from semgrep_agent import semgrep
+from semgrep_agent.constants import ERROR_EXIT_CODE
+from semgrep_agent.constants import FINDING_EXIT_CODE
+from semgrep_agent.constants import NO_RESULT_EXIT_CODE
 from semgrep_agent.exc import ActionFailure
 from semgrep_agent.meta import generate_meta_from_environment
 from semgrep_agent.meta import GithubMeta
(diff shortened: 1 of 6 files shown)


brendongo (approved):

src/semgrep_agent/main.py
```diff
@@ -395,7 +396,7 @@ def protected_main(
             f"| to see your findings in the app, go to {publish_url}/manage/findings?repo={meta.repo_name}"
         )
 
-    exit_code = 1 if blocking_findings and not audit_mode else 0
+    exit_code = 0 if audit_mode else (2 if errors else 1 if blocking_findings else 0)
```
chmccreery: I don't know how standard this is, but might be nice to document with a comment at the top of FindingSets where exit_code is defined.

src/semgrep_agent/semgrep.py
```diff
@@ -274,7 +293,8 @@ def invoke_semgrep(
             for c in chunk:
                 args.append(c)
 
-            _ = semgrep_exec(*args, _timeout=timeout, _err=debug_echo)
+            exit_code = semgrep_exec(*args, _timeout=timeout, _err=debug_echo).exit_code
+            max_exit_code = max(max_exit_code, exit_code)
```
chmccreery, line 297: I would also document this "max of all" behavior with a comment where exit_code is defined at the top of FindingSets

chmccreery (approved):

src/semgrep_agent/main.py
nbrahms: added some self-documenting code

Issue #346, opened 2021-08-06
generateSarif | INPUT_GENERATESARIF causes crash


mschwager:
Background: https://r2c-community.slack.com/archives/CK86BJ5DW/p1628266506046400

When running with SARIF output, `semgrep_agent` crashes under certain repositories.

To repro:

- Visit https://github.com/fleetdm/fleet/tree/93a10e6f943f260e03a73d079b2f6074f659f8a5
- Run `git clone https://github.com/fleetdm/fleet.git`
- `cd fleet` and run `semgrep_agent`
- First, notice the following doesn't crash: `python -m semgrep_agent --config p/kubernetes`
- Second, notice the following does crash: `INPUT_GENERATESARIF=1 python -m semgrep_agent --config p/kubernetes`

```
=== looking for current issues in 1857 files
| Semgrep exited with 1 error:
|    charts/fleet/templates/deployment.yaml:8 Could not parse deployment.yaml as yaml

semgrep-core message:

	Other parsing error: Expected a valid scalar, got
| 6 current issues found
| No ignored issues found
=== not looking at pre-existing issues since all files with current issues are newly created
=== re-running scan to generate a SARIF report
An unexpected error occurred
ERROR:root:Expecting value: line 1 column 1 (char 0)
Traceback (most recent call last):
  File "~/r2c/semgrep-action/src/semgrep_agent/main.py", line 176, in main
    protected_main(**locals())
  File "~/r2c/semgrep-action/src/semgrep_agent/main.py", line 355, in protected_main
    timeout=(timeout if timeout > 0 else None),
  File "~/r2c/semgrep-action/src/semgrep_agent/semgrep.py", line 356, in scan
    timeout=timeout,
  File "~/r2c/semgrep-action/src/semgrep_agent/semgrep.py", line 266, in get_findings
    rewrite_sarif_file(sarif_path)
  File "~/r2c/semgrep-action/src/semgrep_agent/semgrep.py", line 98, in rewrite_sarif_file
    sarif_results = json.load(sarif_file)
  File "~/.pyenv/versions/3.7.0/lib/python3.7/json/__init__.py", line 296, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "~/.pyenv/versions/3.7.0/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
[...]

PR #351, opened 2021-08-20
Prevent semgrep output from breaking sarif output

brendongo:
This should solve https://github.com/returntocorp/semgrep-action/issues/346 I think the issue is that semgrep is printing out non sarif logging code.

### DIFF
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -251,7 +251,6 @@ def get_findings(
                     )
 
     if os.getenv("INPUT_GENERATESARIF"):
-        # FIXME: This will crash when running on thousands of files due to command length limit
         click.echo("=== re-running scan to generate a SARIF report", err=True)
         sarif_path = Path("semgrep.sarif")
         with targets.current_paths() as paths:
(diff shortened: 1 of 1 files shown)


mschwager:
If you try the repro steps here does it fix the error? https://github.com/returntocorp/semgrep-action/issues/346#issue-962920821

Should we add `--quiet` to the `--json` path too?

brendongo:
The json path actually writes the json to a file and reads that so does not need this. The correct way to do this for sarif would be to do that too but figured this was a quick fix

brendongo:
Ugh no you're right it's definitely the --includes taking up too many characters. Gonna do this the right way then

brendongo:
@mschwager the new changes run cleanly and produce a sarif file in the repo you linked in the issue

DrewDennison (approved):

src/semgrep_agent/semgrep.py
```diff
@@ -258,12 +254,12 @@ def get_findings(
         # FIXME: This will crash when running on thousands of files due to command length limit
```
mschwager: We can remove this `FIXME` now :+1:

mschwager (approved):
Nice!

PR #354, opened 2021-08-24
Add explicit info about # of inventory findings and rule findings

### LINKED ISSUE #320: "Rule Recommendation" results included in non-blocking findings count
When using the "Rule Recommendation" functionality of Semgrep App, the CI runner displays very confusing messages like `2 current issues found` and `2 non-blocking findings hidden`.

raghavjain3:
Closes https://github.com/returntocorp/semgrep-action/issues/320

### DIFF
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -187,8 +187,20 @@ def get_findings(
                 for e in findings.errors:
                     for s in render_error(e):
                         click.echo(f"|    {s}", err=True)
+            inventory_findings_len = 0
+            for finding in findings.current:
+                if "__r2c-internal-cai" in finding.check_id:
+                    inventory_findings_len += 1
             click.echo(
                 f"| {unit_len(findings.current, 'current issue')} found", err=True
+            )
+            click.echo(
+                f"| {unit_len(range(len(findings.current)-inventory_findings_len), 'current rule issue')}",
+                err=True,
+            )
+            click.echo(
+                f"| {unit_len(range(inventory_findings_len), 'current inventory issue')}",
+                err=True,
             )
             click.echo(
                 f"| {unit_len(findings.ignored, 'ignored issue')} found",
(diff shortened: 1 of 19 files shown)


src/semgrep_agent/semgrep.py
```diff
@@ -187,9 +187,21 @@ def get_findings(
...
                         click.echo(f"|    {s}", err=True)
+            inventory_findings_len = 0
+            for finding in findings.current:
+                if "__r2c-internal-cai" in finding.check_id:
+                    inventory_findings_len += 1
             click.echo(
                 f"| {unit_len(findings.current, 'current issue')} found", err=True
             )
+            click.echo(
+                f"| {unit_len(range(len(findings.current)-inventory_findings_len), 'current rule issue')}",
```
brendongo, line 198: Why is range necessary?

src/semgrep_agent/semgrep.py
```diff
@@ -187,9 +187,21 @@ def get_findings(
                 for e in findings.errors:
                     for s in render_error(e):
                         click.echo(f"|    {s}", err=True)
+            inventory_findings_len = 0
+            for finding in findings.current:
```
brendongo, line 191: Might be useful to factor into a function to count cai findings given list of findings

src/semgrep_agent/semgrep.py
raghavjain3: unit_len requires an iterable as a parameter so unit_len(int) doesn't work but unit_len(range(int)) does

brendongo (approved):

tests/acceptance/baseline-local-file/new-agent.err
```diff
@@ -10,6 +10,8 @@
 | found 2 files in the paths to be scanned
 === looking for current issues in 2 files
 | 1 current issue found
+| 1 current rule issue
```
dlukeomalley, line 13: What is current rule issue vs. a current issue?

tests/acceptance/baseline-local-file/new-agent.err
```diff
@@ -10,6 +10,8 @@
...
+| No current inventory issues
```
dlukeomalley, line 14: What are inventory issues? HMW make it clear issue doesn't imply "problem" in this case, like it's not something for a user to go fix?

I wonder if we just signal that rule recommendation rules ran, and avoid any specific counts that confuse users or they think they've specifically configured.

chmccreery: Big +1 from me

raghavjain3: Will create a PR today to fix this :)

PR #355, opened 2021-08-25
updated ci output

raghavjain3:


src/semgrep_agent/main.py
```diff
@@ -385,17 +385,14 @@ def protected_main(
...
-            err=True,
-        )
-        click.echo(
-            f"| {unit_len(range(inventory_findings_len), 'non-blocking inventory finding')}",
+            f"| {unit_len(range(len(non_blocking_findings)-inventory_findings_len), 'non-blocking finding')} hidden in output",
             err=True,
         )
+        if inventory_findings_len > 0:
+            click.echo(
+                f"| Inventory scanned since Rule Reccomendations are enabled",
```
chmccreery: nit, but I think this is better grammar 😆 
```suggestion
                f"| Rule recommendations are enabled",
```

chmccreery (commented):
This is probably annoying of me, but to me it makes more sense to count non-inventory findings instead of counting inventory findings and then subtracting them from the total non-blocking findings, especially if we are not showing or ever caring about the number of inventory findings itself. Would it be too much work to rewrite it that way?

src/semgrep_agent/semgrep.py
```diff
@@ -91,15 +91,19 @@ def stats(self) -> Dict[str, Any]:
         }
 
 
-def rewrite_sarif_file(sarif_output: Dict[str, Any], sarif_path: Path) -> None:
+def rewrite_sarif_file(sarif_path: Path) -> None:
     """Fix SARIF errors in semgrep output and pretty format the file."""
+
+    with sarif_path.open() as sarif_file:
```
chmccreery: is this an unrelated change?

src/semgrep_agent/semgrep.py
```diff
@@ -188,16 +192,14 @@ def get_findings(
...
-            click.echo(
-                f"| {unit_len(range(len(findings.current)-inventory_findings_len), 'current rule issue')}",
-                err=True,
-            )
-            click.echo(
-                f"| {unit_len(range(inventory_findings_len), 'current inventory issue')}",
+                f"| {unit_len(range(len(findings.current)-inventory_findings_len), 'current issue')} found",
                 err=True,
             )
+            if inventory_findings_len > 0:
```
chmccreery: maybe we just get rid of this print entirely?

src/semgrep_agent/semgrep.py
raghavjain3: very much so. not sure what happened here but ill fix it

src/semgrep_agent/semgrep.py
raghavjain3: so no reference to scanning inventory at all?

chmccreery: Well doesn't line 391 of main.py already mention it? I think rule recommendations and scanning inventory are kind of the same thing to a user (and all they see on the dashboard UI is an ability to turn rule recommendations on/off)

src/semgrep_agent/semgrep.py
```diff
@@ -188,16 +188,14 @@ def get_findings(
...
+                click.echo(
+                    f"| Inventory detected since Rule Reccomendations are enabled",
```
chmccreery:
```suggestion
                    f"| Scanning inventory because rule recommendations are enabled",
```

src/semgrep_agent/main.py
```diff
@@ -385,17 +385,14 @@ def protected_main(
...
-            err=True,
-        )
-        click.echo(
-            f"| {unit_len(range(inventory_findings_len), 'non-blocking inventory finding')}",
+            f"| {unit_len(range(len(non_blocking_findings)-inventory_findings_len), 'non-blocking finding')} hidden in output",
             err=True,
         )
+        if inventory_findings_len > 0:
+            click.echo(
+                f"| Rule recommendations are enabled",
```
chmccreery:
```suggestion
                f"| Detected technologies for rule recommendation engine",
```

chmccreery (approved):

PR #366, opened 2021-09-07
Remove publishDeployment

chmccreery:
This environment variable adds unnecessary complexity to our config files and has confused many a user. 

First of all, users do not intuitively know what their deployment ID is. Additionally, the name "publishDeployment" is not even the same as "deploymentId" so there is inconsistent terminology making this even more confusing.

The token is a secret, and is sufficient (from a security perspective) to both identify an org and authenticate access to that org. Let's delete publish-deployment once and for all 💪 

I ended up deleting one of our tests instead of updating it to make it pass because it seemed redundant with our live test of the action, and also like it wasn't testing much other than the Mock functionality.


src/semgrep_agent/main.py
```diff
@@ -67,28 +67,28 @@ def url(string: str) -> str:
...
-    show_default="detected from CI env",
-    help="Only show findings introduced since this Git ref",
+    help="Only show findings introduced since this Git ref. If not specified, will be detected from CI environment.",
 )
 @click.option(
     "--publish-token",
     envvar=["INPUT_PUBLISHTOKEN", "SEMGREP_APP_TOKEN"],
     type=str,
-    help="Your semgrep.dev API token (only needed if specifying a publish deployment)",
+    help="Your secret semgrep.dev API token (only needed if using semgrep app)",
```
```suggestion
    help="Your secret semgrep.dev API token (only needed if using Semgrep App)",
```

src/semgrep_agent/main.py
```diff
@@ -234,7 +239,7 @@ def protected_main(
         to_server = "" if publish_url == "https://semgrep.dev" else f" to {publish_url}"
         click.echo(
             get_aligned_command(
-                "manage", f"logged in{to_server} as deployment #{sapp.deployment_id}"
+                "manage", f"logged in{to_server} as {sapp.deployment_name}"
```
```suggestion
                "manage", f"logged in{to_server} as org '{sapp.deployment_name}'"
```

Optional change, to communicate that you authenticate as an organization, not as a user. Might be more confusing this way though.

src/semgrep_agent/semgrep_app.py
```diff
@@ -45,22 +45,43 @@ class Scan:
 class Sapp:
     url: str
     token: str
-    deployment_id: int
+    deployment_id: int = -1
+    deployment_name: str = ""
```
underyx: Maybe it'd be better to initialize this and the ID to None? That way the type checker will remind us everywhere that these might not be available.

underyx (approved):

tests/test_semgrep_app.py
```diff
@@ -1,64 +0,0 @@
...
-    results.findings.new = []
-    results.findings.ignored = []
-    results.findings.searched_paths = []
-
-    sapp.report_results(results, [], [])
-
-    # Check Stdout
-    out, err = capfd.readouterr()
-
-    assert err == f"Server returned following warning: {FAIL_MESSAGE}\n"
```
underyx, line 64: What's up with this test being gone? Seems unrelated.

src/semgrep_agent/main.py
chmccreery: Yeah haha I actually did this first and it looks really weird when it says "as org chmccreery" or "as org returntocorp". But I hear you... maybe if we went all the way and said "signed into the organization returntocorp" that would be best?

tests/test_semgrep_app.py
chmccreery: This change breaks this test: we can no longer call Sapp(...) to instantiate a sapp instance unless we have access to the real app, because it tries to make an API call to the actual semgrep app to get deployment id and name, and the token is just a placeholder ("AAAAAAA"). Only work around I could think of was to mock Sapp() but this test already is basically testing a bunch of mocked versions of things, so I thought maybe it just isn't very useful. We already test the semgrep action live which seems like a much better way to catch issues here 🤷

PR #373, opened 2021-09-17
fix: handle empty sarif scan

brendongo:
### Security

- [ ] Change has security implications (if so, ping the security team)


brendongo:
https://github.com/returntocorp/semgrep-action/pull/374/checks?check_run_id=3636559187 is also failing this test for some reason. Going to force this in to unblock user. This seems like a test error.

PR #389, opened 2021-10-07
Add support for GitHub annotations as formatter output

directionless:
This is the `semgrep_agent` version of https://github.com/returntocorp/semgrep/pull/4001 It adds the [github annotation output format](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-a-notice-message) to the output. 

I wasn't sure what the ideal set of outputs was. Since everything shares standard out, I made this an optional, additional,  output when not using gitlab or json. I kinda wanted to add a `=== findings` heading, but that changes all the test data. 

Probably the best way to test is to make a docker container, and try it in a repo somewhere

### Security

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -343,14 +343,9 @@ def protected_main(
 
     new_findings = results.findings.new
     errors = results.findings.errors
-
-    # Additional github annotation output
-    if github_output:
-        click.echo("=== github annotation output", err=True)
-        click.echo("\n".join([f.to_github() for f in new_findings]))
+    blocking_findings = {finding for finding in new_findings if finding.is_blocking()}
 
     click.echo("=== findings", err=True)
-    blocking_findings = {finding for finding in new_findings if finding.is_blocking()}
     if json_output:
         # Output all new findings as json
         json_contents = [f.to_dict(omit=set()) for f in new_findings]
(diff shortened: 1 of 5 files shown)


src/semgrep_agent/main.py
```diff
@@ -348,6 +358,11 @@ def protected_main(
         }
         click.echo(json.dumps(gitlab_contents))
     else:
+        # Additional github annotation output
+        if github_output:
+            click.echo("=== github annotation output", err=True)
```
brendongo, line 363: Can you add a comment with a link to a doc with the github annotation spec

src/semgrep_agent/main.py
```diff
@@ -120,6 +120,14 @@ def url(string: str) -> str:
 @click.option(
     "--json", "json_output", envvar="SEMGREP_JSON_OUTPUT", hidden=True, is_flag=True
 )
+@click.option(
+    "--github/--no-github",
+    "github_output",
+    help="Additionally output findings in github annotation format",
+    envvar="SEMGREP_GITHUB_OUTPUT",
+    default=os.getenv("GITHUB_ACTIONS") == "true",
```
brendongo, line 128: I think we shouldn't use this env var since people using this semgrep action with the webapp will then get multiple comments. Can we default false.

src/semgrep_agent/main.py
```diff
@@ -348,6 +358,11 @@ def protected_main(
         }
         click.echo(json.dumps(gitlab_contents))
     else:
+        # Additional github annotation output
```
brendongo, line 361: Also might make sense to not have this be mutually exclusive with json_output? So maybe move outside the else.

brendongo (changes requested):

brendongo:
Is there an easy way to format the annotation comment in a nice way? Currently we have a nice html/markdown github comment with extra information when the comment comes from the webap. I'd be worried about a poor developer experience without some of this extra info (how to silence a finding, link to the rule, etc)

brendongo:
Looks like no further progress was made on this. Closing. Feel free to reopen if want to pick this up again.

PR #395, opened 2021-10-10
Use baseline scans for both muted and unmuted issues

chmccreery:
We were not treating muted issues as issues that should get removed via baseline scanning, but this results in confusing behavior where a new (muted) issue shows up in the findings panel with an unassociated PR.

Fixes PD-674

### Security


raghavjain3 (approved):
great catch!

PR #399, opened 2021-10-11
Ignore .semgrep.yml by default

awh-jlr:
Ran into an issue with a scan creating false positives on semgrep's own `.semgrep.yml` file, so let's ignore it by default in addition to the existing `.semgrep/` folder ignore.

### Security

### DIFF
--- a/src/semgrep_agent/templates/.semgrepignore
+++ b/src/semgrep_agent/templates/.semgrepignore
@@ -20,5 +20,6 @@ test/
 tests/
 *_test.go
 
-# Semgrep rules folder
-.semgrep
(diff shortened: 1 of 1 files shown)


brendongo (approved):

underyx:
Hey @awh-jlr, thanks for the contribution! I see that tests are now failing because we were probably expecting a .semgrep.yml hit before in an unrelated test case. This assertion should be safe to update. Could you update the tests by regenerating snapshots as seen here? https://github.com/returntocorp/semgrep-action/blob/develop/CONTRIBUTING.md#tests

brendongo:
Closing as stale

PR #400, opened 2021-10-12
doc(PRIVACY): Add "timing metrics" item; Clarify "data not collected"

underyx:
### Security


ievans (approved):
Very nice changes, super @underyx

PR #403, opened 2021-10-15
Add --gitlab-secrets-json

chasemp:
Gitlab has moved to a separate scanner type and schema for secret detection.  These results are reported and stored distinct from SAST findings.  The current secret detection schema is reflected at 14.0.0. 

* Note CVE is meaningful in terms of determining unique findings here
* Additional scanner information is included compared to SAST
* Added start and end times for reporting back to gitlab
* Tested on gitlab SaaS offering using generic.secrets.security.detected-aws-access-key-id-value

### Security

### DIFF
--- a/src/semgrep_agent/findings.py
+++ b/src/semgrep_agent/findings.py
@@ -166,18 +166,11 @@ class Finding:
 
     def to_gitlab_secrets(self) -> Dict[str, Any]:
 
-        # IDs need to be persistent across findings for the same issue
-        # This needs to account for the same rule matching in different
-        # files and the same file
-        # file_path + check_id + matching lines = uniqueness
-        id_hash = sha224(
-            self.path.encode("utf-8")
-            + self.check_id.encode("utf-8")
-            + str(self.line).encode("utf-8")
-            + str(self.end_line).encode("utf-8")
-        ).hexdigest()
         return {
-            "id": id_hash,
+            # IDs need to be persistent across findings for the same issue
+            # This needs to account for the same rule matching in different
+            # files and the same file -- even on the same syntactic value
+            "id": str(self.syntactic_identifier_int()),
             "category": "secret_detection",
             # CVE is a required field from Gitlab schema.
             # It also is part of the determination for uniqueness
--- a/src/semgrep_agent/findings.py
+++ b/src/semgrep_agent/findings.py
@@ -163,6 +163,46 @@ class Finding:
             ],
         }
 
+    def to_gitlab_secrets(self) -> Dict[str, Any]:
+        id = str(uuid.uuid5(uuid.NAMESPACE_URL, str(self.path)))
+        return {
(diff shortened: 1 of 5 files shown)


chasemp:
Screen Shot 2021-10-15 at 2 42 52 PM
Screen Shot 2021-10-15 at 2 43 01 PM

src/semgrep_agent/findings.py
```diff
@@ -163,6 +164,59 @@ def to_gitlab(self) -> Dict[str, Any]:
             ],
         }
 
+    def to_gitlab_secrets(self) -> Dict[str, Any]:
+
+        # IDs need to be persistent across findings for the same issue
+        # This needs to account for the same rule matching in different
+        # files and the same file
+        # file_path + check_id + matching lines = uniqueness
+        id_hash = sha224(
```
brendongo: Does the syntactic_identifier_int work for this? it's a hash of check_id, path,

brendongo (changes requested):
Can you also add a test like https://github.com/returntocorp/semgrep-action/blob/develop/tests/test_gitlab_output.py intention is to catch any unintended changes if we refactor findings.py or something.

src/semgrep_agent/findings.py
chasemp: I think yes, I'm going to do some testing to make sure.  good call.

src/semgrep_agent/findings.py
```diff
@@ -163,6 +164,52 @@ def to_gitlab(self) -> Dict[str, Any]:
             ],
         }
 
+    def to_gitlab_secrets(self) -> Dict[str, Any]:
+
+        return {
+            # IDs need to be persistent across findings for the same issue
+            # This needs to account for the same rule matching in different
+            # files and the same file -- even on the same syntactic value
+            "id": str(self.syntactic_identifier_int()),
```
brendongo: You might want to use syntactic_identifier_str if it needs to be a string.

```suggestion
            "id": self.syntactic_identifier_str(),
```

chasemp:
Made this a draft while I revalidate after recent changes.  Cheers and thanks for your help.

chasemp:
I ran through various test scenarios:

- Same finding same line
- Different findings same line
- Different Findings different line

With treatment (dismiss, et al).  Seems to work a treat.  Good stuff.

chasemp (commented):
test added :)

tests/test_gitlab_secret_output.py
```diff
@@ -0,0 +1,16 @@
+import json
+from pathlib import Path
+
+from jsonschema import validate
+
+
+def test_gitlab_secret_output(run_semgrep_agent, get_test_root):
```
brendongo, line 7: 🔥 👍

brendongo (approved):

PR #430, opened 2021-11-18
feat(semgrep): call with timeout-threshold

brendongo:
If a file times out with semgrep on 3 rules, it will no longer
send the file for analysis with other rules since we will likely just keep timing out

### Security

### DIFF
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -277,7 +277,7 @@ def _get_head_findings(
             "--dryrun",
             "--time",
             "--timeout-threshold",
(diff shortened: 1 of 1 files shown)


src/semgrep_agent/semgrep.py
```diff
@@ -276,6 +276,8 @@ def _get_head_findings(
             "--autofix",
             "--dryrun",
             "--time",
+            "--timeout-threshold",
+            "5",
```
brendongo:
```suggestion
            "3",
```

emjin (approved):

PR #431, opened 2021-11-18
Fix failing when running with --config auto

brendongo:
User was reporting semgrep-action failing with JSONDecodeError when doing a baseline scan. Locally semgrep was working fine. User was running with semgrep-action --config auto.

Failure was due to semgrep aborting with auto was used with metrics off.

### DIFF
--- a/src/semgrep_agent/semgrep.py
+++ b/src/semgrep_agent/semgrep.py
@@ -394,8 +394,8 @@ def _update_baseline_findings(
                         *config_args,
                     ]
 
-                    # Can only disable metrics if auto is not a passed config
-                    # disable so only one metrics run per semgrep-action run
+                    # If possible, disable metrics so that we get metrics only once per semgrep-action run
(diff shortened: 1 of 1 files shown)


DrewDennison (approved):

src/semgrep_agent/semgrep.py
```diff
@@ -390,11 +390,15 @@ def _update_baseline_findings(
...
                         "--disable-nosem",
                         "--json",
-                        "--metrics",
-                        "off",  # only count one semgrep run per semgrep-agent run
                         *extra_args,
                         *config_args,
                     ]
+
+                    # Can only disable metrics if auto is not a passed config
+                    # disable so only one metrics run per semgrep-action run
```
emjin:
```suggestion
                    # If possible, disable metrics so that we get metrics only once per semgrep-action run
                    # However, if run with config auto we must allow metrics to be sent
```

emjin (approved):

Issue #429, opened 2021-11-17
Re-running scan to generate SARIF report times out


sim642:
This recently started happening to all of our semgrep workflows, e.g. https://github.com/goblint/analyzer/runs/4236663683?check_suite_focus=true#step:4:44.
As the log shows, semgrep first runs fine and finds no issues. Then it starts re-running scan to generate SARIF report and that for some reason does nothing but waste 30min, time out and thus fail without having generated it.

brendongo:
Hi @sim642 are you able to run with setting the SEMGREP_AGENT_DEBUG env flag to 1

sim642:
Sure! The workflow run with debugging enabled was here: https://github.com/goblint/analyzer/runs/4272519434?check_suite_focus=true#step:4:1. Since it's truncated and I'm not sure if the full thing is accessible for download or not, I'm attaching the full log here: [semgrep-30min.log](https://github.com/returntocorp/semgrep-action/files/7574251/semgrep-30min.log). It ends with the following:
```
2021-11-20T08:35:26.5809042Z === [DEBUG] [1.599  Info       Main.Run_semgrep     ] Analyzing docs/artifact-descriptions/sas21.md
2021-11-20T08:35:26.5809846Z === [DEBUG] [1.599  Info       Main.Run_rules       ] looking for (Analyze_rule.And
2021-11-20T08:35:26.5810396Z === [DEBUG]    [(Analyze_rule.Or
2021-11-20T08:35:26.5810977Z === [DEBUG]        [(Analyze_rule.Idents ["clear"]); (Analyze_rule.Idents ["reset"]);
2021-11-20T08:35:26.5812070Z === [DEBUG]          (Analyze_rule.Idents ["add"]); (Analyze_rule.Idents ["remove"]);
2021-11-20T08:35:26.5812757Z === [DEBUG]          (Analyze_rule.Idents ["replace"])]);
2021-11-20T08:35:26.5813293Z === [DEBUG]      (Analyze_rule.Or
2021-11-20T08:35:26.5813921Z === [DEBUG]         [(Analyze_rule.Idents ["iter"]); (Analyze_rule.Idents ["fold"])])
2021-11-20T08:35:26.5814889Z === [DEBUG]      ]) in docs/artifact-descriptions/sas21.md
2021-11-20T08:35:26.5816347Z === [DEBUG] [1.599  Info       Main.Match_rules     ] checking docs/artifact-descriptions/sas21.md with 1 rules
[...]

PR #437, opened 2021-11-22
Catch error debug_echo

brendongo:
From https://github.com/returntocorp/semgrep-action/issues/429 looks like sh passes bytes object to the _err command. Looks like an exception thrown while handling stderr of a sh command causes us to deadlock somehow maybe because the stderr pipe can never be closed. Just shove it in try catch for now but sh commands should probably call a different function

### Security

### DIFF
--- a/src/semgrep_agent/utils.py
+++ b/src/semgrep_agent/utils.py
@@ -42,7 +42,12 @@ def debug_echo(text: str) -> None:
     else:
         logging.info(text)
         return
-    text = "\n".join(prefix + line for line in text.splitlines())
+    try:
+        modified_text = "\n".join(prefix + line for line in text.splitlines())
+        text = modified_text
+    except Exception as e:
+        logging.info(e)
+
     click.echo(text, err=True)
 
 


src/semgrep_agent/utils.py
```diff
@@ -42,7 +42,12 @@ def debug_echo(text: str) -> None:
     else:
         logging.info(text)
         return
-    text = "\n".join(prefix + line for line in text.splitlines())
+    try:
+        modified_text = "\n".join(prefix + line for line in text.splitlines())
+        text = modified_text
+    except Exception as e:
+        logging.info(e)
```
emjin, line 49: Can you still log the initial error, just without the prefix?

emjin (approved):

src/semgrep_agent/utils.py
emjin: Ah you were suggesting that this can't be outputted at all?

brendongo: It's possible? Haven't dug into how sh is using _out and _err much yet

PR #440, opened 2021-11-24
More debug logs in TargetManager + lazy load submodules

brendongo:
### Security


emjin (approved):

HUMM1984 (commented):
Teat

PR #448, opened 2021-12-07
Fix Semgrep error reporting

nbrahms:
long_msg was removed from core error JSON blobs, breaking error reporting.
Fix by falling back to the "message" field.

Also truncate at new lines to avoid spam.

### Security

### DIFF
--- a/src/semgrep_agent/utils.py
+++ b/src/semgrep_agent/utils.py
@@ -63,7 +63,10 @@ def maybe_print_debug_info(meta: "GitMeta") -> None:
 
 def render_error(error: Mapping[str, Any]) -> Sequence[str]:
     spans = error.get("spans")
-    msg = error.get("long_msg", "")
+    msg: str = error.get("long_msg", error.get("message", ""))
+    newline_ix = msg.find("\n")
+    if newline_ix > -1:
+        msg = msg[: newline_ix - 1]
     if spans:
         return [f"{s['file']}:{s['start']['line']} {msg}" for s in spans]
     return [msg]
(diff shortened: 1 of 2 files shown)


src/semgrep_agent/utils.py
```diff
@@ -63,7 +63,10 @@ def maybe_print_debug_info(meta: "GitMeta") -> None:
 
 def render_error(error: Mapping[str, Any]) -> Sequence[str]:
     spans = error.get("spans")
-    msg = error.get("long_msg", "")
+    msg: str = error.get("long_msg", error.get("message", ""))
```
brendongo, line 66: If long_msg was removed should we just not even try to get it anymore?

src/semgrep_agent/utils.py
```diff
@@ -63,7 +63,10 @@ def maybe_print_debug_info(meta: "GitMeta") -> None:
...
+    newline_ix = msg.find("\n")
+    if newline_ix > -1:
+        msg = msg[: newline_ix - 1]
```
brendongo, line 72: Why are we doing this modification?

src/semgrep_agent/utils.py
emjin: long_msg is still used in some cases (rule parsing)

nbrahms: ^^ yup

src/semgrep_agent/utils.py
nbrahms: b/c otherwise we'd dump enormous but useless stack traces to the console

brendongo: Can you add an inline comment mentioning that

brendongo (approved):

src/semgrep_agent/utils.py
brendongo: To help know what needs moving later when we port to semgrep-cli

PR #456, opened 2021-12-14
Get correct files for diff

### LINKED ISSUE #454: Merge commits causing extra findings
Semgrep is reporting findings in files not touched. The relevant PR includes a large merge commit which may be the reason for this.

emjin:
Closes https://github.com/returntocorp/semgrep-action/issues/454

Asana was seeing semgrep flag findings that were introduced in a different branch later merged into this one. This is because we were running `git diff` without `--merge-base`. `--merge-base` (to my understanding) is the difference between just straight comparing two branches and getting the diffs you see in PRs. However, there are lots of edge cases, so please check for yourself: https://git-scm.com/docs/git-merge-base

The added test tries to mimic the scenario Asana seemed to have: two branches branch off of main, then branch A introduces a vulnerability and gets merged into main, then branch B merges main with the vulnerability. Branch B should not flag branch A's vuln.

We also needed to fix an existing test. My understanding is: the existing test deletes a file containing a vulnerability from main but does not push it. The new branch thus still includes the file with the vulnerability. When we run semgrep-agent on the new branch, it should not find that vulnerability, even though it is no longer on main.

### Security


brendongo (approved):

tests/acceptance/mutlimerge-generic/commands.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - -c
+      - git checkout all-good && git merge main
+  - name: new log
+    description: git log with commits
+    command: git log --stat
+    returncode: 0
+    expected_out: new-log.out
+    expected_err: new-log.err
+  - name: new agent
+    description: expect no errors even though merge contains an error
```
emjin, line 34: Note: I indeed validated that without merge-base, this flags the error introduced by introduce-err, which is not the branch (all-good) being scanned.

Issue #421, opened 2021-11-08
Unable to upload "semgrep.sarif" as it is not valid SARIF


roisec:
Hi,
I am got this error for the github action:

Run github/codeql-action/upload-sarif@v1
Uploading results
Error details: instance requires property "version"
Error details: instance requires property "runs"
Error: Unable to upload "semgrep.sarif" as it is not valid SARIF:

Have any ideas?

ievans:
Hi @RoiSec, can you share the SARIF file?

ievans:
Probably a dupe of https://github.com/returntocorp/semgrep-action/issues/378

roisec:
https://github.com/RoiSec/caspion/runs/4141290747

brendongo:
@RoiSec doesnt look like the sarif file is there?

brendongo:
I think closed by #457

Issue #378, opened 2021-09-21
Invalid (empty) SARIF file generated


nelenkov:
The scan action generates an invalid SARIF file, causing the upload action to fail.

Maybe this is caused by this recent change: https://github.com/returntocorp/semgrep-action/commit/9e14dce842a906eec042c6c6383735cddfe20f96

It appears that Github is doing some validation on the SARIF file.

Output is below:

```
 using semgrep rules from https://semgrep.dev/c/p/phpcs-security-audit
| using path ignore rules from .semgrepignore
| not on head ref 09770910...; checking that out now...
| reporting findings introduced by these commits:
|   * 0977091 ....
| looking at No changed paths
| found No files in the paths to be scanned
=== looking for current issues in No files
| No current issues found
| No ignored issues found
| returning to original head revision efe2d174...
=== not looking at pre-existing issues since there are no current issues
=== re-running scan to generate a SARIF report
=== exiting with success status
1s
Run github/codeql-action/upload-sarif@v1
Uploading results
Error details: instance requires property "version"
Error details: instance requires property "runs"
Error: Unable to upload "semgrep.sarif" as it is not valid SARIF:
- instance requires property "version"
- instance requires property "runs"
Error: Unable to upload "semgrep.sarif" as it is not valid SARIF:
- instance requires property "version"
- instance requires property "runs"
    at validateSarifFileSchema (/home/runner/work/_actions/github/codeql-action/v1/lib/upload-lib.js:196:15)
    at uploadFiles (/home/runner/work/_actions/github/codeql-action/v1/lib/upload-lib.js:256:9)
    at Object.uploadFromActions (/home/runner/work/_actions/github/codeql-action/v1/lib/upload-lib.js:132:18)
    at async run (/home/runner/work/_actions/github/codeql-action/v1/lib/upload-sarif-action.js:49:29)
    at async runWrapper (/home/runner/work/_actions/github/codeql-action/v1/lib/upload-sarif-action.js:61:9)
```

ultras5631:
same issue there. any fixes?

andyjones:
I'm hitting this as well when the PR being scanned hasn't changed any files that semgrep would scan (eg. readme or test files). #421 looks like the same issue as it says "found No files in the paths to be scanned" so it generates a serif file that only contains `{}`. 

Github validate the serif file against https://github.com/github/codeql-action/blob/main/src/sarif_v2.1.0_schema.json

brendongo:
[...]

PR #457, opened 2021-12-14
workaround #378: generate a valid sarif file when no files are scanned

andyjones:
I use semgrep in github actions and upload the sarif file to Github. The job sometimes fails with a schema validation error when no files are scanned as described in #378 and #421 

This tweaks the output to generate a minimum sarif file that validates against the 2.1.0 json schema.

### Security

### DIFF
--- a/tests/unit/test_sarif.py
+++ b/tests/unit/test_sarif.py
@@ -1,0 +1,28 @@
+import json
+from pathlib import Path
+from typing import Any
+from typing import Dict
+
+import pytest
+from jsonschema import validate
+
+from semgrep_agent.semgrep import rewrite_sarif_file
+
+
+@pytest.fixture
+def sarif_schema():
+    with (Path(__file__).parent / "sarif-schema-2.1.0.json").open() as fd:
+        schema = json.load(fd)
+    return schema
+
+
+def test_rewrite_empty_sarif_file(tmp_path: Path, sarif_schema):
+    sarif_output: Dict[str, Any] = {}
+    sarif_path = tmp_path / "semgrep.sarif"
+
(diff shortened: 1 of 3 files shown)


brendongo:
Sarif spec copied here matches https://github.com/oasis-tcs/sarif-spec/blob/master/Documents/CommitteeSpecifications/2.1.0/sarif-schema-2.1.0.json but looks like that spec doesn't match the spec published on https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317736 specifically the suppression object has `state` property in the former but is `status` in the later. Going to copy over the spec from https://docs.oasis-open.org/sarif/sarif/v2.1.0/cos02/schemas/sarif-schema-2.1.0.json

tests/unit/test_sarif.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@pytest.fixture
+def sarif_schema():
+    with (Path(__file__).parent / "sarif-schema-2.1.0.json").open() as fd:
+        schema = json.load(fd)
+    return schema
+
+
+def test_rewrite_empty_sarif_file(tmp_path: Path, sarif_schema):
```
brendongo, line 19: Thanks for the test!

brendongo (approved):
Thanks for the contribution @andyjones !

andyjones:
thanks @brendongo , I missed that! FYI in case it helps others, github use a slightly different/more lax schema when they validate the upload as well: https://github.com/github/codeql-action/commits/main/src/sarif_v2.1.0_schema.json

PR #459, opened 2021-12-16
Remove semgrepignore

mmcqd:
This PR removes `.semgrepignore` functionality from semgrep-action, as it can now be handled entirely by the Semgrep CLI. It also removes the target/directory expansion functionality. Previously we recorded the number of files found via target expansion and the number of files ignored via `click.echo`. This should probably be added back into the CLI. Depends on
https://github.com/returntocorp/semgrep/pull/4542 to explicitly set a semgrepignore file that overrides any existing `.semgrepignore`, so we can include ignores from semgrep-app.

### Security


mmcqd:
Test failing because we no longer report files scanned/ignored counts. Going to update the test expectations.

tests/acceptance/baseline-generic/new-agent.err
```diff
@@ -3,11 +3,9 @@
 | manage            - not logged in
 === setting up agent configuration
 | using semgrep rules from https://semgrep.dev/c/p/r2c-ci
```
emjin, line 5: Can you run locally and check the logs (in `.semgrep_logs`) just to make sure that the "using semgrepignore" information is still being logged somewhere? Preferably it also logs the "skipping 1 file based on path ignores" information.

src/semgrep_agent/ignores.py
```diff
@@ -1,312 +0,0 @@
...
-        if not this_path.exists():
-            return
-
-        if this_path.is_file():
-            yield Entry(this_path, self._survives(this_path))
-        else:
-            for e in os.scandir(this_path_str):
-                if e.is_symlink():
-                    continue
-                elif this_path.match(LOG_FOLDER):
```
emjin, line 117: Deleting this means we're going to now scan the `.semgrep_logs` folder that semgrep-action generates. I think we need to port this to semgrep. Make sure when testing to run on newly cloned repos, but also to run semgrep-action _twice_.

mmcqd: Can we just add `.semgrep_logs` to the default `.semgrepignore`?

emjin: ummmm sure

(Hesitance is because most users will probably always want to ignore it, but better to have it all in one place)

mmcqd: added to default `.semgrep_ignore`

tests/acceptance/baseline-generic/new-agent.err
mmcqd: Yes this gets logged by Semgrep, so it shows up in `.semgrep_logs/semgrep_agent_logs`

emjin (dismissed):
Can't merge yet :( Online ignores not accounted for

test_app_ignore.py
```diff
@@ -0,0 +1,3 @@
+# This should be ignored by semgrep-app. If this file is ever caught when running semgrep-action
+# it is an indication that semgrep-action is not repsecting ignore rules from semgrep-app
+x = 5 == 5
```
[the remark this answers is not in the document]
mmcqd: This is, correctly, indicating that we are failing to include ignores from semgrep-app in the semgrepignore file we pass to the CLI. Once Semgrep is released, this should be solved.

brendongo (approved):

PR #465, opened 2021-12-17
Save the config file used when downloading from the app

emjin:
### Security

### DIFF
--- a/src/semgrep_agent/main.py
+++ b/src/semgrep_agent/main.py
@@ -46,6 +46,7 @@ ENV_VAR_HELP_TEXT = "\n        ".join(
 )
 
 LOG_FILE = LOG_FOLDER + "/semgrep_agent_logs"
+SEMGREP_RULES_FILE = LOG_FOLDER + "/semgrep_app_rules.yaml"
 
 
 def url(string: str) -> str:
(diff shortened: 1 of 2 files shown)


src/semgrep_agent/semgrep_app.py
```diff
@@ -156,10 +156,10 @@ def fetch_rules_text(self) -> str:
...
 
     def download_rules(self) -> Tuple[Path, List[str], List[str]]:
-        """Save the rules configured on semgrep app to a temporary file"""
-        # hey, it's just a tiny YAML file in CI, we'll survive without cleanup
-        rules_file = tempfile.NamedTemporaryFile(suffix=".yml", delete=False)  # nosem
-        rules_path = Path(rules_file.name)  # nosem
+        """Save the rules configured on semgrep app to a file in .semgrep_logs"""
+        """so that it persists for debugging """
+        rules_file = SEMGREP_RULES_FILE
+        rules_path = Path(rules_file)  # nosem
```
underyx: I assume the nosem is not needed since we're not using `tempfile`'s `.name` anymore.

```suggestion
        rules_path = Path(rules_file)
```

brendongo (approved):

src/semgrep_agent/main.py
```diff
@@ -325,6 +326,7 @@ def protected_main(
             click.secho(message, err=True, fg="red")
             sys.exit(1)
         config = (str(local_config_path),)
+        logging.info(f"Config saved to { SEMGREP_RULES_FILE }")
```
```suggestion
        logging.info(f"Config saved to {SEMGREP_RULES_FILE}")
```

I think black might be broken if it's not enforcing this style :/

PR #507, opened 2022-02-02
Add test that .semgrep_logs is ignored, even if not present in .semgrepignore

mmcqd:
semgrep-action should always ignore `.semgrep_logs`, even if the user has a custom .semgrepignore that does not list it.

### Security


tests/acceptance/ignore-logs/new-agent.err
```diff
@@ -0,0 +1,13 @@
+| versions          - semgrep x.y.z on Python x.y.z
+| environment       - running in environment git, triggering event is 'unknown'
+| manage            - not logged in
+=== setting up agent configuration
+| using semgrep rules from https://semgrep.dev/c/p/r2c-ci
+| reporting findings introduced by these commits:
+|   * aaaaaaa add finding
+| looking at 2 changed paths
+=== looking for current issues in 2 files
```
emjin, line 9: Why is this 2 files? I just thought it was always 1 file now

emjin: Or is that only for full scans

mmcqd: I'm pretty sure that's only for full scans, where we just pass the whole repo to the CLI. Diff scans have to pass specific individual files.

tests/acceptance/ignore-logs/commands.yaml
```diff
@@ -0,0 +1,21 @@
+steps:
```
emjin, line 1: Maybe note that the agent output in this case is a little wonky ("looking at 2 changed paths" and "looking for current issues in 2 files" but then ignoring those files)

emjin (approved):
Looks good

brendongo (approved):

PR #511, opened 2022-02-09
Remove .semgrepignore handling

underyx:
The .semgrepignore logic is left up to the Semgrep CLI from now on.

The agent needs to do two things to make this possible:
 1. Add default --exclude patterns, such as test directories
 2. Pass web UI-configured ignores through as --exclude patterns

Closes PA-732

### Security


tests/acceptance/symlink-dir/new-agent.err
```diff
@@ -5,10 +5,7 @@
 | using semgrep rules from https://semgrep.dev/c/p/r2c-ci
 | reporting findings introduced by these commits:
 |   * aaaaaaa Add symlink to dir
-| Skipping bar/foo since it is a symlink to a directory: foo
```
underyx, line 8: does semgrep CLI handle symlinks like the agent used to?

underyx: silly question — that's exactly what this test checks

PR #513, opened 2022-02-12
Use semgrep --config policy

brendongo:
### Security


underyx (approved):

ievans:
@brendongo didn't we land on calling this "app" instead?

underyx:
**thought (non-blocking)**: Rules not longer locked in at scan start
The one major consideration to note is that we no longer lock the rules in use to what the policies had at the start of scan. This is a very short margin of time for race conditions, but it's still there. It might cause errors during ingestion or on the frontend if the scan gets a finding from a rule that was not supposed to be part of the scan according to policy.

underyx:
**question**: Does --config=policy add inventory rules the same way?
It might be worth verifying that there's no difference in the logic of whether inventory rules should run or not, between the old SCAN_ID/rules.yaml endpoint, and what the CLI now uses.

brendongo:
@ievans yep. will just change towards the end when the dust settles :D

brendongo:
re race condition: I think I can look into the delta of using the scan endpoint since I'm moving the rest of semgrep_app.py anyway

inventory: good point. I have a seperate env var for that and I did forget to use it https://github.com/returntocorp/semgrep/blob/develop/semgrep/semgrep/config_resolver.py#L576 instead of modifying this code to add that env var I'll just merge this as is and if modifying to use the scan endpoint to freeze the rules doesnt work, use the env var

PR #517, opened 2022-02-16
Promote autofix in PR comments

chmccreery:
Instead of hiding autofix in PR comments behind a secret environment variable, users can now specify in the app whether they want to enable autofix on a per-repo basis.

I took this opportunity to also refactor how policy names are stored, since returning them from report_start made less sense than keeping them as a property inside of the scan. These changes are backwards-compatible, so can go in whenever.

### Security

### DIFF
--- a/src/semgrep_agent/semgrep_app.py
+++ b/src/semgrep_agent/semgrep_app.py
@@ -42,6 +42,8 @@ RETRYING_ADAPTER = requests.adapters.HTTPAdapter(
 class Scan:
     id: int = -1
     ignore_patterns: List[str] = field(default_factory=list)
+    policy_list: str = "unknown"
+    autofix: bool = False
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


underyx (approved):

dlukeomalley:
@chmccreery Do we need any back compat here? I suspect the answer is "no" as it's been experimental.

chmccreery:
> @chmccreery Do we need any back compat here? I suspect the answer is "no" as it's been experimental.

@dlukeomalley this is backwards compatible

src/semgrep_agent/semgrep_app.py
```diff
@@ -130,9 +132,10 @@ def report_start(self, meta: GitMeta) -> str:
             self.scan = Scan(
                 id=glom(body, T["scan"]["id"]),
                 ignore_patterns=glom(body, T["scan"]["meta"].get("ignored_files", [])),
+                policy_list=glom(body, T["policy"]),
+                autofix=glom(body, T.get("autofix", False)),
```
chmccreery, line 136: Default of False is for backwards compatibility

PR #556, opened 2022-04-05
edit generate of id. https://github.com/returntocorp/semgrep-action/i…

zakrush:
…ssues/488

### Security


brendongo:
Whoops I did some clean up of old branches and it closed this PR.

@zakrush latest code does not parse semgrep output and instead just relies on semgrep to output the desired format. I think you can reopen the same pr on the semgrep repo if this functionality is necessary

PR #566, opened 2022-04-14
Add multiple values to envvars

### LINKED ISSUE #565: Semgrep-action no longer works with multiple configs
When I pass multiple `--config` flags to semgrep-agent, it used to use all the rules, but now it only uses the last ruleset given. When I pass `--config "r/php" --config "r/javascript"`, I expect it to run both PHP and JavaScript rules, but it only uses JavaScript rules.

```
$ docker run -v "$PWD:/mysrc" -w "/src" returntocorp/semgrep-agent:v1 semgrep-agent --config "r/php" --config "r/javascript"
=== Running: SEMGREP_RULES=r/javascript semgrep ci
Running `semgrep ci` without API token but with configs ('r/javascript',)
...
```

Previously, it would show something like this:

```
| versions          - semgrep 0.81.0 on Python 3.9.10
| environment       - running in environment git, triggering event is 'unknown'
| using semgrep rules from https://semgrep.dev/c/r/php
| using semgrep rules from https://semgrep.dev/c/r/javascript
...
```

Sjord:
When passing --config multiple times, put the multiple values space-separated in the environment variable.

https://click.palletsprojects.com/en/8.1.x/options/#multiple-values-from-environment-values

Closes #565

### Security

### DIFF
--- a/src/semgrep_agent.py
+++ b/src/semgrep_agent.py
@@ -55,7 +55,10 @@ class ForwardAction(argparse.Action):
     def __call__(self, _, namespace, values, option_string=None) -> None:  # type: ignore
         envvar = FLAG_TO_ENV.get(option_string)
         if envvar:
-            os.environ[envvar] = values
+            if envvar in os.environ:
+                os.environ[envvar] += " " + values
+            else:
+                os.environ[envvar] = values
 
         new_flag = FLAG_TO_FLAG.get(option_string)
         if new_flag:
--- a/src/semgrep_agent.py
+++ b/src/semgrep_agent.py
@@ -132,7 +132,7 @@ def run_sarif_scan() -> None:
         """
     )
 
-    envvars = [f"{k}={v} " for k, v in os.environ.items() if k in ENV_VARS_TO_LOG]
+    envvars = [f'{k}="{v}" ' for k, v in os.environ.items() if k in ENV_VARS_TO_LOG]
     print(
         "=== Running: " + "".join(envvars) + " ".join(cmd),
         file=sys.stderr,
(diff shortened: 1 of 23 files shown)


src/semgrep_agent.py
```diff
@@ -55,7 +55,10 @@ def __init__(self, option_strings, dest, **kwargs) -> None:  # type: ignore
     def __call__(self, _, namespace, values, option_string=None) -> None:  # type: ignore
         envvar = FLAG_TO_ENV.get(option_string)
         if envvar:
-            os.environ[envvar] = values
+            if envvar in os.environ:
+                os.environ[envvar] += " " + values
```
underyx, line 60: Sorry about the oversight of missing this, and thank you for fixing it! I worry though that we might have some users who have set both a flag and an env var for the same option, and this would break their CI jobs now 🤔 

Specifically, imagine if someone committed this (silly) configuration: `SEMGREP_APP_TOKEN=123 semgrep ci --publish-token=123`. It'd previously work correctly, and would break after this as token would be set to `"123 123"`.

I think I'm comfortable with setting a condition where we have this behavior only if the envvar is `SEMGREP_RULES`. Care to update the PR to do that instead? Otherwise, I could just fix up your PR and merge.

Sjord: I have added a commit that restricts the multiple values to `--config` and `--audit-on`. I included `--audit-on` since that also has `multiple=True` in [semgrep ci.py](https://github.com/returntocorp/semgrep/blob/develop/semgrep/semgrep/commands/ci.py#L136).

underyx: Thanks so much :heart:

underyx (approved):

PR #567, opened 2022-04-20
Release 0.89.0

brendongo:
### Security

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Upcoming - Date
 
+# 2022-04-20
+
+## Changed
+
+- Use semgrep 0.89.0
+
 # 2022-03-24
 
 ## Changed
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM returntocorp/semgrep:0.87.0
+FROM returntocorp/semgrep:0.89.0
 
 USER root
 WORKDIR /semgrep-agent
--- a/tests/acceptance/semgrep-rules-repo/local-config-full-scan.out
+++ b/tests/acceptance/semgrep-rules-repo/local-config-full-scan.out
@@ -5,7 +5,6 @@ Findings:
      python.lang.best-practice.arbitrary-sleep
         time.sleep() call; did you mean to leave this in?
 
-
          12┆ t.sleep(5)
           ⋮┆----------------------------------------
          14┆ t.sleep(0.1)
--- a/tests/acceptance/semgrep-rules-repo/local-config-some-new-results-audit-mode.out
+++ b/tests/acceptance/semgrep-rules-repo/local-config-some-new-results-audit-mode.out
@@ -7,7 +7,6 @@ Findings:
         vulnerability where translators could include malicious script tags in their translations.
         Consider using `force_escape` to explicitly escape a transalted text.
 
-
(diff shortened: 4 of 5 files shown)


underyx (approved):

sudomann:
This release is broken. Our jobs don't work with this one because:
```
$ docker run --rm -v "${PWD}:/src" --workdir /src returntocorp/semgrep-agent:v1 semgrep-agent --config https://semgrep.dev/p/r2c --config https://semgrep.dev/p/secrets --config /src/utilities/scripts/semgrep.yml --baseline-ref develop

=== Running: SEMGREP_RULES=/src/utilities/scripts/semgrep.yml SEMGREP_BASELINE_COMMIT=develop semgrep ci
Running `semgrep ci` without API token but with configs ('/src/utilities/scripts/semgrep.yml',)
Scan environment:
  versions    - semgrep 0.89.0 on python 3.10.4
  environment - running in environment git, triggering event is unknown

Fetching configuration from semgrep.dev
Cannot find a commit with reference 'develop'. Possible reasons:

- the referenced commit does not exist
- the current working directory is not a git repository
- the git binary is not available

Try running `git show develop` to debug the issue.
```

The branch certainly exists and is visible to git. It worked just fine when I used an older version.

underyx:
@sudomann sorry about that 😯 unfortunately we cannot reproduce the error, and we don't see failed scans that started occurring with 0.89.0 for logged in users.

Could you share the CI job that causes this failure?

I also have a guess about what might be the issue 🤔 Maybe this recent git release is blocking git commands from being executed in your CI: https://github.blog/2022-04-12-git-security-vulnerability-announced/

If that is the case, a possible fix would be changing your command to

```
docker run --rm -v "${PWD}:/src" --workdir /src returntocorp/semgrep-agent:v1 /bin/sh -c "git config --global --add safe.directory /src && semgrep-agent --config https://semgrep.dev/p/r2c --config https://semgrep.dev/p/secrets --config /src/utilities/scripts/semgrep.yml --baseline-ref develop"
```

underyx:
Otherwise, you can probably get an error message from git by running

```
docker run --rm -v "${PWD}:/src" --workdir /src returntocorp/semgrep-agent:v1 git show develop
```

sudomann:
I see, you're right, thanks!
```
Status: Downloaded newer image for returntocorp/semgrep-agent:v1
fatal: unsafe repository ('/src' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /src
```

I got it to work with: `docker run --rm -v "${PWD}:/src" --workdir /src returntocorp/semgrep-agent:v1 bash -c "git config --global --add safe.directory /src && git show develop"` so I guess the images released today have a newer version of git with this mitigation?

underyx:
@sudomann that's right! The git version was implicitly bumped today since we last released.

underyx:
I wonder if we can do anything to make it so our users don't need to worry about the git configuration, it really sucks that you have to add that extra command :(

But as far as I can tell if we did anything to sidestep this new security feature, we'd just be reintroducing the git vulnerability.

underyx:
And thanks a lot for bringing this to our attention! I added a note about this to today's changelog https://github.com/returntocorp/semgrep-action/blob/develop/CHANGELOG.md#2022-04-20

PR #581, opened 2022-05-25
chore: Bump semgrep to 0.94.0

underyx:
### Security


nmote (approved):
Looks reasonable, but check the boxes in the summary (assuming they should be checked)

PR #601, opened 2022-07-06
Remove --json deprecation warning

underyx:
### Security


brendongo:
Why are we removing the warning?

brendongo (approved):

PR #602, opened 2022-07-07
update semgrep-action with a bump script for conformity

spencerdrak:
### Security


nbrahms (approved):
Reviewing regexes is so difficult :)

PR #649, opened 2022-10-14
Update tests

brendongo:
### Security
- [ ] Changelog has been updated
- [ ] Change has no security implications (otherwise, ping the security team)

PR #653, opened 2022-10-20
Revert "Update tests"

underyx:
returntocorp/semgrep-action#649 removed some (but not all) tests that assumed the registry will have useless-eqeq in p/ci

this reverts the removal and fixes these tests to explicitly run the useless-eqeq rule


emjin:
So what happened exactly? I'm now even more confused

IagoAbal (approved):

PR #662, opened 2022-11-08
chore: Add bump_version workflow

zyannes:
### Security

### DIFF
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -1,0 +1,87 @@
+name: bump-version
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version of semgrep to use"
+        required: true
+        type: string
+
+jobs:
+  bump-version:
+    runs-on: ubuntu-20.04
+    permissions:
+      id-token: write
+      contents: write
+      pull-requests: write
+      checks: write
+    env:
+      NEW_SEMGREP_VERSION: ${{ github.event.inputs.version }}
+      ACTOR: ${{ github.actor }}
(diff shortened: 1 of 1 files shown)


.github/workflows/bump_version.yml
```diff
@@ -0,0 +1,84 @@
...
+        run: |
+          # check if the branch already has a pull request open
+          if gh pr list --head ${SOURCE} | grep -vq "no pull requests"; then
+              # pull request already open
+              echo "pull request from SOURCE ${SOURCE} to TARGET ${TARGET} is already open";
+              echo "cancelling release"
+              exit 1
+          fi
+          # open new pull request with the body of from the local template.
+          res=$(gh pr create --title "${TITLE}" --body "Bump Semgrep Version to ${NEW_SEMGREP_VERSION}" \
```
spencerdrak: can we also just remove the output capture here? makes it easier to debug if anything goes wrong.

spencerdrak (approved):
LGTM - will the `semgrep-ci` bot need additional perms to be able to run this? 

I suspect yes, and I'll work with chris to adjust that.

.github/workflows/bump_version.yml
zyannes: Good call! Updated in latest commit

zyannes:
@spencerdrak The `semgrep-ci` bot will need to have `checks: write` and `pull_requests: write` permissions. I suspect it already has these permissions since this workflow performs similar API calls as the `release-action` workflow. I'll do a simple test once this PR gets merged in to check if the existing permissions are sufficient.

